PR #1, opened 2021-11-17
feat(Arbitration): standard update

unknownunknown1:
The document describing the updated standard https://docs.google.com/document/d/1VIBspauNpw7A8ZeHaSXN2t4RLyGdtognpZMvSk1cJPE/edit

### DIFF
--- a/contracts/src/arbitration/IArbitrator.sol
+++ b/contracts/src/arbitration/IArbitrator.sol
@@ -7,11 +7,11 @@ import "./IArbitrable.sol";
 /**
  * @title Arbitrator
  * Arbitrator interface for CourtV2.
- * This interface for the most part follows the ERC-792 standard but also allows the appeal crowdfunding on arbitrator's side.
+ * This interface follows the ERC-792 standard but also allows the appeal crowdfunding on arbitrator's side.
  * When developing arbitrator contracts we need to:
  * - Define the functions for dispute creation (createDispute). Don't forget to store the arbitrated contract and the disputeID (which should be unique, may nbDisputes).
  * - Define the function for appeal crowdfunding (fundAppeal) in order to appeal the ruling.
- * - Define the functions for cost display (arbitrationCost and appealCost).
+ * - Define the functions for cost display (arbitrationCost and fundingStatus).
  * - Allow giving rulings. For this a function must call arbitrable.rule(disputeID, ruling).
  */
 interface IArbitrator {
@@ -59,12 +59,17 @@ interface IArbitrator {
     function arbitrationCost(bytes memory _extraData) external view returns (uint256 cost);
 
     /**
-     * @dev Compute the cost of an appeal.
-     * @param _disputeID ID of the dispute to be appealed.
-     * @param _extraData Can be used to give additional info of the dispute.
-     * @return cost Required appeal cost.
+     * @dev Return the funded amount and funding goal for one (or the subset) of choices.
+     * Note that the status info may not be available if the dispute is not currently appealable.
+     * @param _disputeID The ID of the dispute to appeal.
+     * @param _choices The one (or the subset) of choices to check the funding status of.
+     * @return funded The amount funded so far for this subset in wei.
+     * @return goal The amount to fully fund this subset in wei.
      */
-    function appealCost(uint256 _disputeID, bytes memory _extraData) external view returns (uint256 cost);
+    function fundingStatus(uint256 _disputeID, uint256[] memory _choices)
+        external
+        view
+        returns (uint256 funded, uint256 goal);
 
     /**
      * @dev Compute the start and end of the dispute's appeal period, if possible. If appeal is impossible: should return (0, 0).
@@ -76,7 +81,6 @@ interface IArbitrator {
 
     /**
      * @dev Make a contribution to fund one (or the subset) of possible choices in order to appeal the ruling.
-     * Note that the desired appeal system will be defined by the selected Dispute Kit.
      * @param _disputeID The ID of the dispute to appeal.
      * @param _choices The choices to contribute to.
      */
--- a/contracts/src/arbitration/IArbitrator.sol
+++ b/contracts/src/arbitration/IArbitrator.sol
@@ -6,20 +6,14 @@ import "./IArbitrable.sol";
 
 /**
  * @title Arbitrator
- * Arbitrator interface for CourtV2.
+ * Arbitrator interface that implements the new arbitration standard.
  * Unlike the ERC-792 this standard doesn't have anything related to appeals, so each arbitrator can implement an appeal system that suits it the most.
  * When developing arbitrator contracts we need to:
  * - Define the functions for dispute creation (createDispute). Don't forget to store the arbitrated contract and the disputeID (which should be unique, may nbDisputes).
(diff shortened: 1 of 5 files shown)


contracts/src/arbitration/IArbitrator.sol
```diff
@@ -0,0 +1,98 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8;
+
+import "./IArbitrable.sol";
+
+/**
+ * @title Arbitrator
+ * Arbitrator interface for CourtV2.
+ * This interface for the most part follows the ERC-792 standard but also allows the appeal crowdfunding on arbitrator's side.
```
jaybuidl, line 10: "follows ~~for the most part~~ the ERC-792 standard" 

I suggest that we either:
1. We amend the ERC-792 specification so that this interface complies with it - but the specs must deal with v1 compatibility. The amendment can be made outside [the ERC-1 process](https://eips.ethereum.org/EIPS/eip-1#eip-process) because the status is "pre-draft" and not formally tracked.
2. Or we start a new ERC spec which leaves kleros-v1 compatibility untouched at the expense of having to rebuild legitimacy for this new ERC from scratch.

contracts/src/arbitration/IArbitrator.sol
```diff
@@ -0,0 +1,98 @@
...
+    function createDispute(uint256 _choices, bytes calldata _extraData) external payable returns (uint256 disputeID);
+
+    /**
+     * @dev Compute the cost of arbitration. It is recommended not to increase it often, as it can be highly time and gas consuming for the arbitrated contracts to cope with fee augmentation.
+     * @param _extraData Can be used to give additional info on the dispute to be created.
+     * @return cost Required cost of arbitration.
+     */
+    function arbitrationCost(bytes memory _extraData) external view returns (uint256 cost);
+
+    /**
```
jaybuidl, line 61: This is now an internal function of the Arbitrator which does not need standardization so it should be possible to remove it just like appeal() has been removed.

contracts/src/arbitration/IArbitrator.sol
```diff
@@ -0,0 +1,98 @@
...
+    function appealPeriod(uint256 _disputeID) external view returns (uint256 start, uint256 end);
+
+    /**
+     * @dev Make a contribution to fund one (or the subset) of possible choices in order to appeal the ruling.
+     * Note that the desired appeal system will be defined by the selected Dispute Kit.
+     * @param _disputeID The ID of the dispute to appeal.
+     * @param _choices The choices to contribute to.
+     */
+    function fundAppeal(uint256 _disputeID, uint256[] calldata _choices) external payable;
+
```
jaybuidl, line 84: A view function showing the funding status, funding goal and number of required funded choices would be useful. It's different from the old `appealCost()` which did not account for the funders reward. Maybe something like:
```solidity
    /**
     * @dev Return the funded amount and funding goal for this choice, as well as the minimum number of choices which must be funded to obtain an appeal. The status info may not be available if the dispute is not currently appealable.
     * @param _disputeID The ID of the dispute to appeal.
     * @param _choice The choice corresponding to this status.
     * @return funded The amount funded so far for this choice in wei.
     * @return goal The amount to fully fund this choice in wei.
     * @return minNbOfChoicesToFund The number of choices which must be fully funded to obtain an appeal.
     */
    function fundingStatus(uint256 _disputeID, uint256 _choice)
        external
        view
        returns (
            uint256 funded,
            uint256 goal,
            uint256 minNbOfChoicesToFund
        );
```

jaybuidl (changes requested):

jaybuidl:
Interface changes tentatively:
image

contracts/src/arbitration/IArbitrator.sol
unknownunknown1: Personally,I was thinking that it'll be a new ERC

contracts/src/arbitration/IArbitrator.sol
```diff
@@ -0,0 +1,98 @@
...
+     */
+    function appealPeriod(uint256 _disputeID) external view returns (uint256 start, uint256 end);
+
+    /**
+     * @dev Make a contribution to fund one (or the subset) of possible choices in order to appeal the ruling.
+     * Note that the desired appeal system will be defined by the selected Dispute Kit.
+     * @param _disputeID The ID of the dispute to appeal.
+     * @param _choices The choices to contribute to.
+     */
+    function fundAppeal(uint256 _disputeID, uint256[] calldata _choices) external payable;
```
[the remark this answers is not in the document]
unknownunknown1: I left a comment in the doc. In short, it might get too tricky if we have subsets

contracts/src/arbitration/IArbitrator.sol
```diff
@@ -0,0 +1,98 @@
...
+     * @dev Compute the start and end of the dispute's appeal period, if possible. If appeal is impossible: should return (0, 0).
+     * @param _disputeID ID of the dispute.
+     * @return start The start of the period.
+     * @return end The end of the period.
+     */
+    function appealPeriod(uint256 _disputeID) external view returns (uint256 start, uint256 end);
+
+    /**
+     * @dev Make a contribution to fund one (or the subset) of possible choices in order to appeal the ruling.
+     * Note that the desired appeal system will be defined by the selected Dispute Kit.
```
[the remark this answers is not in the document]
unknownunknown1: Alright, makes sense

contracts/src/arbitration/IArbitrator.sol
unknownunknown1: See my comment in the doc. 
uint256 minNbOfChoicesToFund - Is it possible to have more than two choices though? Even if fund a certain subset it'll be always funded as a countermeasure to the first funded choices that should always be separated, according to the yellow paper. So even with the subset funded choices are de-facto just 2

contracts/src/arbitration/IArbitrator.sol
```diff
@@ -0,0 +1,98 @@
...
+     */
+    function arbitrationCost(bytes memory _extraData) external view returns (uint256 cost);
+
+    /**
+     * @dev Compute the cost of an appeal.
+     * @param _disputeID ID of the dispute to be appealed.
+     * @param _extraData Can be used to give additional info of the dispute.
+     * @return cost Required appeal cost.
+     */
+    function appealCost(uint256 _disputeID, bytes memory _extraData) external view returns (uint256 cost);
```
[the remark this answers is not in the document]
unknownunknown1: Agreed

contracts/src/arbitration/IArbitrator.sol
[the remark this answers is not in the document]
jaybuidl: Just read your comments, I agree it's not trivial actually. Let's not worry about this.

contracts/src/arbitration/IArbitrator.sol
jaybuidl: Indeed that's more consistent: `uint256 _choice` --> `uint256[] _choices`

jaybuidl (approved):

jaybuidl (approved):

PR #6, opened 2021-12-10
Add the native arbitrum bridge

shalzz:


.gitignore
```diff
@@ -10,6 +10,7 @@ yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
 .pnpm-debug.log*
+.yarn
```
jaybuidl, line 13: this section of the file is autogenerated, the change should be before or after

shalzz: Are we using yarn zero-installs?
This is the offending line that made me add this: https://github.com/kleros/kleros-v2/blob/feature/bridge/.gitignore#L186

If we are not then this line needs to be commented out

shalzz: Ideally, I'd say not to commit generated and downloaded file, as then tend to usually cause merge conflicts when they can anyways be generated/downloaded by the build system

jaybuidl (approved):

PR #7, opened 2021-12-10
Add arbitrum specific gateway contracts

shalzz:
This is the native bridge implementation.

Some unsolved problems:
* Transferring ETH for paying the arbitrationCost after bridging:
  1.) This is inherently more complex than just state bridging (involves a bunch of error handling and refund handling which is very bridge specific and may not be refund claimable from non-EOA accounts)
  2.) This is especially true for the xdai bridge since eth<->xdai transfer is not intuitive

* Calculation of the arbitration cost in relation to the numOfJurors:
  1.) The same logic for this from the KlerosCore would need to be re-implemented here. 
  2.) This would most likely require a governance function that updates the jurorFees per court on the gateways whenever there is a governance proposal to so on the KlerosCore contract as well.
For now, the arbitration cost is just a constant i.e it does not scale with the numOfJurors.


jaybuidl:
Shaleen there is no need for such complex inheritance, in particular diamond inheritance is generally an anti-pattern which may yield [unexpected result due to the C3 linearization](https://forum.openzeppelin.com/t/solidity-diamond-inheritance/2694) (a call to a super can be substituted by a call to a sibling).

Favor [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance): BaseXGateway does not need to inherit from ILXBridge, from an OO point of view, a gateway **_has_** a bridge seems more correct than a gateway **_is_** a bridge. So ILXBridge can just be a contract variable passed during initialization of BaseXGateway.

shalzz:
Well, this is not diamond inheritance

jaybuidl:
```
        IL1Bridge
        /       \
ArbL1Bridge   BaseForeignGateway
        \       /
     EthereumGateway
```

Am I missing anything?

shalzz:
> ```
>         IL1Bridge
>         /       \
> ArbL1Bridge   BaseForeignGateway
>         \       /
>      EthereumGateway
> ```
> 
> Am I missing anything?

`IL1Bridge` is an interface. So I wouldn't really count that as an inherited contract.

jaybuidl:
Right IL1Bridge is an interface. I still think that it can simplified bearing in mind that a Gateway _is not_ a Bridge, it breaks the abstraction of the bridge which the gateway provides. To me BaseForeignGateway _has_ a IL1Bridge. In this case the gateway would still need to expose the getSubmissionPrice() function but it can be arranged.

That would give:
```
 IL1Bridge ⬥-------- BaseForeignGateway
     ▲                       ▲
     |                       |
ArbL1Bridge           EthereumGateway
```

shalzz:
The implementation would definitely change while solving the open problems I mentioned above which are bigger issues IMO.

contracts/src/gateway/BaseForeignGateway.sol
```diff
@@ -0,0 +1,126 @@
...
+
+        disputeID = localDisputeID++;
+        bytes32 disputeHash = keccak256(
+            abi.encodePacked(
+                chainID,
+                blockhash(block.number - 1),
+                "createDispute",
+                disputeID,
+                arbitrationCost(_extraData),
+                _extraData
```
jaybuidl, line 58: Let's add the msg.sender to the disputeHash, as the gateway may receive disputes from other contracts later.

jaybuidl (approved):
Not feature complete but good enough to integrate and test the existing features.

PR #11, opened 2022-01-10
Feat/kleros liquid proxy

fnanni-0:


contracts/src/kleros-v1/KlerosV1Governor.sol
```diff
@@ -0,0 +1,190 @@
...
+     *  @param _disputeID The ID of the dispute as defined in KlerosLiquid.
+     */
+    function relayDispute(uint256 _disputeID) external {
+        require(klerosLiquidDisputeIDtoGatewayDisputeID[_disputeID] == 0, "Dispute already relayed");
+        IKlerosLiquid.Dispute memory KlerosLiquidDispute = klerosLiquid.disputes(_disputeID);
+        (uint256[] memory votesLengths, , uint256[] memory totalFeesForJurors, , , ) = klerosLiquid.getDispute(
+            _disputeID
+        );
+
+        // Check that no juror was yet drawn. Add a function to finalize the juror drawing and move the dispute to vote.
```
shalzz, line 81: I guess this is still a TODO?

contracts/src/kleros-v1/KlerosV1Governor.sol
```diff
@@ -0,0 +1,190 @@
...
+     */
+    function changeGovernor(address _governor) external onlyByGovernor {
+        governor = _governor;
+    }
+
+    /** @dev Relays disputes from KlerosLiquid to Kleros v2. Only disputes in the evidence period of the initial round can be realyed.
+     *  @param _disputeID The ID of the dispute as defined in KlerosLiquid.
+     */
+    function relayDispute(uint256 _disputeID) external {
+        require(klerosLiquidDisputeIDtoGatewayDisputeID[_disputeID] == 0, "Dispute already relayed");
```
shalzz, line 75: I don't see any specification where we don't allow disputeID to be zero. The current KlerosLiquid has a dispute with id 0: https://court.kleros.io/cases/0

This is the same behaviour that I have in the foreign gateway where disputeID starts from 0 and increments.

I can modify the gateway contract to start the disputeID with 1 but is this really required?
(This contract could instead use a boolean mapping)

contracts/src/kleros-v1/KlerosV1Governor.sol
```diff
@@ -0,0 +1,190 @@
...
+
+        klerosLiquid.executeGovernorProposal(address(this), totalFeesForJurors[0], "");
+
+        uint256 minJurors = votesLengths[0];
+        bytes memory extraData = abi.encode(KlerosLiquidDispute.subcourtID, minJurors);
+        uint256 arbitrationCost = foreignGateway.arbitrationCost(extraData);
+        require(totalFeesForJurors[0] >= arbitrationCost, "Fees not high enough."); // If this doesn't hold at some point, it could be a big issue.
+        uint256 gatewayDisputeID = foreignGateway.createDispute(KlerosLiquidDispute.numberOfChoices, extraData);
+        klerosLiquidDisputeIDtoGatewayDisputeID[_disputeID] = gatewayDisputeID;
+        require(gatewayDisputeID != 0, "ID must be greater than 0.");
```
shalzz, line 93: See the above comment about disputeID

contracts/src/kleros-v1/KlerosV1Governor.sol
```diff
@@ -0,0 +1,190 @@
...
+        IKlerosLiquid.Dispute memory KlerosLiquidDispute = klerosLiquid.disputes(_disputeID);
+        (uint256[] memory votesLengths, , uint256[] memory totalFeesForJurors, , , ) = klerosLiquid.getDispute(
+            _disputeID
+        );
+
+        // Check that no juror was yet drawn. Add a function to finalize the juror drawing and move the dispute to vote.
+        require(KlerosLiquidDispute.period == IKlerosLiquid.Period.evidence, "Invalid dispute period.");
+        require(votesLengths.length == 1, "Cannot relay appeals.");
+
+        klerosLiquid.executeGovernorProposal(address(this), totalFeesForJurors[0], "");
```
shalzz, line 85: Note, the gateway contracts require paying the arbitration cost on the same chain. So this might be counter-intuitive

contracts/src/kleros-v1/KlerosV1Governor.sol
```diff
@@ -0,0 +1,190 @@
...
+
+        uint256 minJurors = votesLengths[0];
+        bytes memory extraData = abi.encode(KlerosLiquidDispute.subcourtID, minJurors);
+        uint256 arbitrationCost = foreignGateway.arbitrationCost(extraData);
+        require(totalFeesForJurors[0] >= arbitrationCost, "Fees not high enough."); // If this doesn't hold at some point, it could be a big issue.
+        uint256 gatewayDisputeID = foreignGateway.createDispute(KlerosLiquidDispute.numberOfChoices, extraData);
```
shalzz, line 91: This will revert unless the arbitrationCost is payed

contracts/src/kleros-v1/KlerosV1Governor.sol
```diff
@@ -0,0 +1,190 @@
...
+
+        bytes4 functionSelector = IArbitrable.rule.selector;
+        bytes memory data = abi.encodeWithSelector(functionSelector, dispute.klerosLiquidDisputeID, _ruling);
+        klerosLiquid.executeGovernorProposal(klerosLiquidDispute.arbitrated, 0, data);
+    }
+
+    /** @dev Registers jurors' tokens which where locked due to relaying a given dispute. These tokens don't count as locked.
+     *  @param _disputeID The ID of the dispute as defined in KlerosLiquid.
+     */
+    function notifyFrozenTokens(uint256 _disputeID) external {
```
shalzz, line 124: When should this function be called?

If after a dispute is successfully relayed then probably we can call it from the `relayDispute` function itself.

contracts/src/kleros-v1/KlerosV1Governor.sol
jaybuidl: See L82: if the dispute is not in Period.evidence it reverts. So dispute id 0 will not be relayed.

contracts/src/kleros-v1/KlerosV1Governor.sol
jaybuidl: @shalzz how about having createDispute() return disputeHash instead? There is no requirement for the disputeID to be sequential and the disputeHash won't ever be exposed to the legacy arbitrables or arbitrator.

contracts/src/kleros-v1/KlerosV1Governor.sol
jaybuidl: I have doubts about this part too, currently discussing it with Fede.

contracts/src/kleros-v1/KlerosV1Governor.sol
shalzz: Maybe you forgot, the gateway contracts have to follow the arbitrable/arbitator interfaces. since the KlerosV1 proxy will not be the only user of the gateway contracts.

contracts/src/kleros-v1/KlerosV1Governor.sol
fnanni-0: Right, there is no specification about this. The assumption that dispute ID can't be zero comes in handy to save storage. I think it would be nice to see this in the foreign gateway and maybe even add it to the standard.

contracts/src/kleros-v1/KlerosV1Governor.sol
fnanni-0: It should be called:
 if tokensAtStakePerJuror > 0 && jurors were drawn && all jurors were drawn
 
 Note that jurors could get drawn after the dispute is relayed.

contracts/src/kleros-v1/KlerosV1Governor.sol
fnanni-0: Good catch. Should be solved now.

contracts/src/kleros-v1/KlerosV1Governor.sol
jaybuidl: This is not a new idea, it has been proposed by Henrique already a few months ago.

The disputeHash is a bytes32 so obviously it fits in a uint256 which satisfies the Arbitrator interface. As I said it won't be exposed to the v1 contracts. For future v2 integrations talking to the gateways, it will be known in advance not to expect sequential IDs which again is not a requirement from the standard.

shalzz: The disputeHash requires a disputeID as part of the pre-image. Do you mean to remove the disputeID from the pre-image and use the disputeHash as the  disputeID or just overwrite the disputeID with the generated disputeHash?

contracts/src/kleros-v1/KlerosV1Governor.sol
shalzz: Okay, I've updated the foreign gateway contracts to increment the dispute ID from one. That should be enough for now, unless we see any other issues.

contracts/src/kleros-v1/KlerosV1Governor.sol
jaybuidl: Okay no worries let's keep this part as it, now that the gatewayDisputeId starts from 1. 
a9df0ab

jaybuidl (approved):
Should be fine to start integrating with the other branches and start testing. We can iterate from there.

PR #28, opened 2022-01-21
fix(GHSA-64g7-mvw6-v9qj): upgraded shelljs

jaybuidl:


jaybuidl:
Encountered cascading issues including a weird checksum mismatch on a typescript patch. 
image

Tried deleting yarn.lock and reinstalling in vain.
Ended up upgrading both yarn and typescript to the next minor version and it worked...

PR #35, opened 2022-02-01
refactor(bridge): favor composition over inheritance

jaybuidl:
"BaseForeignGateway is an IL1Bridge" replaced by "BaseForeignGateway has an IL1Bridge".

In particular: 
- "arbitrum.EthereumGateway has an ArbL1Bridge" 
- "xdai.EthereumGateway has a xDaiL1Bridge"

Similarly for BaseHomeGateway.

The main differences:
- Access control is required in the implementation of `ILxBridge.sendCrossDomainMessage()`
- Slightly more gas, mostly relevant for Mainnet. In practice this bridge is not used during normal operations later (challenges only), and the fast bridge version will be called for a batch of messages, so the gas cost is incurred for a batch, not per message.

### DIFF
--- a/contracts/src/bridge/IL1Bridge.sol
+++ b/contracts/src/bridge/IL1Bridge.sol
@@ -28,9 +28,9 @@ abstract contract IL1Bridge {
         bytes memory _calldata,
         uint256 _maxGas,
         uint256 _gasPriceBid
-    ) internal virtual returns (uint256);
+    ) public payable virtual returns (uint256);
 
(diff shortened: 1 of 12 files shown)


jaybuidl:
This decision will have mostly an impact on the long term but not so much at this stage. More feedback from other team members would be helpful as there are merits to both approaches.

jaybuidl:
Can be closed now, it has been addressed as part of #36

PR #36, opened 2022-02-04
Partial Fast Bridge

shalzz:
This still needs access control around the fast bridge, but It shouldn't be a blocker for the demo.


jaybuidl (approved):

contracts/src/bridge/FastBridgeSender.sol
```diff
@@ -44,14 +44,30 @@ contract FastBridgeSender is IFastBridgeSender {
...
+    function sendSafe(address _receiver, bytes memory _calldata) external payable {
         // The safe bridge sends the encoded data to the FastBridgeReceiver
         // in order for the FastBridgeReceiver to resolve any potential
         // challenges and then forwards the message to the actual
         // intended recipient encoded in `data`
         // TODO: For this encodedData needs to be wrapped into an
         // IFastBridgeReceiver function.
         // TODO: add access checks for this on the FastBridgeReceiver.
-        safebridge.sendCrossDomainMessage(address(fastBridgeReceiver), encodedData);
+        // TODO: how much ETH should be provided for bridging? add an ISafeBridge.bridgingCost()
```
shalzz, line 74: Bridging from L2 to L1 does not need any bridging cost/fees.

contracts/src/bridge/FastBridgeSender.sol
```diff
@@ -15,31 +15,36 @@ import "./interfaces/IFastBridgeSender.sol";
...
      * via the fast bridge mechanism
      *
-     * TODO: probably needs some access control either on the sender side
-     * or the receiver side
-     *
      * @param _receiver The L1 contract address who will receive the calldata
      * @param _calldata The receiving domain encoded message data.
      */
     function sendFast(address _receiver, bytes memory _calldata) external {
+        require(msg.sender == fastSender, "Access not allowed: Fast Sender only.");
```
shalzz, line 46: The reason I didn't add this initially was to make the fast bridge app-agnostic.

contracts/src/bridge/FastBridgeSender.sol
jaybuidl: Right. Given the existing native bridges, the fees would be required only for L1 to L2, and we are not using the SafeBridge for it right now. So it's more for future proofing.

PR #58, opened 2022-03-17
feat: Fast Bridge unhappy path (obsolete)

### LINKED ISSUE #53: Fast Bridging: unhappy path, no extra hop
A participant makes a `claim(hash)` about a single message hash. Another participant may `challenge(hash)` which requires falling back to the Safe Bridge and waiting a number of days to resolve.

No extra hop for this iteration: we focus on the Safe Bridge from Home chain to Ethereum mainnet where a native bridge is available. Safe Bridging to other foreign chains can be added afterwards, for example Home chain -> Ethereum -> Gnosis chain afterwards.

Only 1 claim per hash is allowed at the moment, regardless of challenge status, further claims for the same hash are ignored.

Bounties can be added later.

The full specs are not public yet.

jaybuidl:
Resolves #53

### DIFF
--- a/contracts/src/bridge/FastBridgeReceiver.sol
+++ b/contracts/src/bridge/FastBridgeReceiver.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 
 /**
- *  @authors: [@shalzz]
+ *  @authors: [@shalzz, @hrishibhat]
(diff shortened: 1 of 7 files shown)


jaybuidl:
Closing this, addressed by #60 instead.

PR #59, opened 2022-03-17
feat: Fast Bridge unhappy path with Merkle tree

### LINKED ISSUE #54: Fast Bridging: message batching
Message batching by making a `claim(hash)` about a Merkle tree root hash of messages instead of the hash of a single message.

`verifyAndRelay()` on the foreign chain must verify the hash given the Merkle tree branch and data for a specific message.

Any `challenge(hash)` requires falling back to the Safe Bridge and waiting a number of days to resolve.

Only 1 claim per hash is allowed at the moment, regardless of challenge status, further claims for the same hash are ignored.

Bounties can be added later.

The full specs are not public yet.

jaybuidl:
Resolves #54

### DIFF
--- a/contracts/src/bridge/FastBridgeReceiver.sol
+++ b/contracts/src/bridge/FastBridgeReceiver.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 
 /**
- *  @authors: [@shalzz*, @hrishibhat]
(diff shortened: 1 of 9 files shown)


jaybuidl:
Has a dependency on issue #53 which has just resolved by PR #60.
We will need to rebase this on top of master manually.

jaybuidl:
Work done in a separate branch `feat/challenge-path-3-with-merkle-root` as it has diverged too much. 
Closing this.

PR #69, opened 2022-04-06
feat: Initial web app for the court

jaybuidl:


AndreiMVP:
I think we should have the file organisation as simple as is convenient. If a component is used multiple times by all means it should be in a separate file. But in case of a component with <50 lines which will be used only once I don't see why it should be in a separate file. It only makes the directory tree more complex and navigating between components harder. This applies to the header and footer related components. I would even say that the layout, header and footer components should be one file.

web/src/index.tsx
```diff
@@ -1,24 +1,20 @@
 import React from "react";
 import { createRoot } from "react-dom/client";
-import { HashRouter } from "react-router-dom";
 import { ThemeProvider } from "styled-components";
 import { GlobalStyle } from "./styles/global-style";
 import { lightTheme } from "./styles/themes";
-import Layout from "components/layout";
+import App from "./app";
+import { BrowserRouter as Router } from "react-router-dom";
```
alcercu, line 7: We should use HashRouter just in case we want to deploy to IPFS in the future https://stackoverflow.com/questions/64546654/how-to-use-react-routing-with-ipfs

alcercu:
Thanks for the refactor @AndreiMVP! Regarding the file organization, I prefer to keep it in separated files even if these components are used once and only in the context of the layout. This is mainly because they might be small now, but once we start adding the desktop styling they might grow. Also, I think it's better to keep them abstracted, so in the future, if someone wants to modify some part of the layout, not related to the header and footer, they can focus directly on the part that matters.

web/src/index.tsx
AndreiMVP: Ok, I removed it because the react-router docs didn't recommend using HashRouter unless you have to and I didn't know ipfs would need that type of router

jaybuidl:
Netlify deploy is successful https://deploy-preview-69--kleros-v2.netlify.app  ✅

jaybuidl:
Let's try merge this branch and continue working in a new more specific one, maybe `feat/cases-page`.

AndreiMVP (approved):
Looks OK for now

alcercu:
@jaybuidl @AndreiMVP I think this is ready to be merged, the last codeclimate duplication issues are in the theme file and in a file (LatestCases) that it's not fetching data yet (so the duplication comes from the manually-inserted data). There are issues regarding some type exporting, I think this is due to the difference between tslint (deprecated) and eslint with typescript support (the one we use locally), we need to look into updating codeclimate's linter.

alcercu (approved):

PR #73, opened 2022-04-21
Feat(web): add Home page

alcercu:


jaybuidl (approved):

jaybuidl:
The sonarcloud alert was on a generated file which is now ignored, but the scanner has not run again.

PR #75, opened 2022-05-10
Feat/merkle library

shotaronowhere:
Implements a space efficient merkle tree data structure and merkle proof functions as stand alone contracts which will be incorporated into the fastBridgeReceiver and fastBridgeSender. 

Contrary to the name of the branch, the merkle tree is not actually implemented in a library.

On rollups, there is no contract size limit, and duplicating the merkle tree data structure and merkle proof functions directly into the fast bridge will save gas. 

- On L1, incorporating the merkle proof functions directly into the fast bridge receiver is a significant gas saving (maybe 3k gas per message).
- On L2, incorporating the merkle tree data structure into the fast bridge sender is a modest gas savings. Discarding the L1 fixed transaction cost and L1 calldata cost from L2 transactions, insertions with a merkle tree library cost 0.000000251911548938 eth versus  0.00000013056669626 eth when the merkle tree data structure is directly implemented in the fast bridge contract. This is a significant relative gas savings, but modest overall (saves fractions of pennies which can add up over time).

The leaves of the merkle tree are hashed with a different hash function (sha256) than the interior leaves to prevent a so-called '2nd pre-image attack'

https://flawed.net.nz/2018/02/21/attacking-merkle-trees-with-a-second-preimage-attack/


jaybuidl:
Required to resolve #54

jaybuidl (approved):

PR #90, opened 2022-06-13
feat(KC): justification + small fixes

unknownunknown1:
Added features:

1. Unstaking of ineligible jurors
2. Justification of the vote
3. Safe transfers to avoid reverts
4. 1 vote per address in PoH DK
5. Changing the default court to General instead of Forking. The latter is basically rudimental now. I need clarifications on how to implement the forking process
6. Some small fixes along the way


jaybuidl:
We could add the `_justification` as part of the commitment hash? [L330](https://github.com/kleros/kleros-v2/pull/90/files#diff-1aa232bc8f244086557d76af53b852ca679d33ebd11aa9085066273e9bfd525aR330)

jaybuidl (approved):

PR #93, opened 2022-06-17
chore: fixed the linters configuration

jaybuidl:
Typescript files were not linted by the pre-commit hook.

Two major changes:
- VS Code is going to use the top-level .prettierrc.json so it may affect the `web` and `subgraph` packages.
- The precommit hook does not run `eslint` anymore for .js files, only `prettier`.

What do you think? @alcercu @AndreiMVP

### DIFF
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,5 +1,5 @@
 {
-  "*.{js,ts}": "prettier --write",
+  "*.{js,jsx,ts,tsx}": "prettier --write",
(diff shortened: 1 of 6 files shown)


alcercu:
hey @jaybuidl!
Just checked and vim is getting the prettier config from the web directory, so no problem on that side. I don't know about VSCode though.
Regarding the pre-commit hook for typescript files, if it's also intended for the web package, we should have the file extension `tsx` in the rules too.

PR #106, opened 2022-06-22
chore: bumping up yarn, typescript, commitling config

jaybuidl:
Moved commitling to a separate config file


jaybuidl:
so borked

Issue #182, opened 2022-08-02
Allow jurors to override a vote commitment


donosonaumczuk:
Basically allow DisputeKit's `castCommit` function to be called multiple times during the commit period, each call overrides the commits of the previous one. This allows a juror to change its mind during the commit phase.

shotaronowhere:
This sounds like a good idea to me.

0xferit:
This feature was forgotten it seems. I recall we wanted to implement it for v2, 2-3 years ago. It has some advantages like making implementing on-chain bribing contracts harder.

donosonaumczuk:
Glad to see you agree with this change. I just opened a PR #190 for it!

PR #190, opened 2022-08-04
feat: allow overriding a vote commitment

donosonaumczuk:
Removes validations from `castCommit` function to allow submitting votes multiple times overriding the previous ones as long as the current period is `Commit`.

This also has the positive side effect of making the `castCommit` function cheaper.

This PR is the implementation for issue #182.

### DIFF
--- a/contracts/src/arbitration/dispute-kits/DisputeKitClassic.sol
+++ b/contracts/src/arbitration/dispute-kits/DisputeKitClassic.sol
@@ -269,7 +269,8 @@ contract DisputeKitClassic is BaseDisputeKit, IEvidence {
         }
     }
 
-    /** @dev Sets the caller's commit for the specified votes.
+    /** @dev Sets the caller's commit for the specified votes. It can be called multiple times during the
+     *  commit period, each call overrides the commits of the previous one.
      *  `O(n)` where
      *  `n` is the number of votes.
      *  @param _coreDisputeID The ID of the dispute in Kleros Core.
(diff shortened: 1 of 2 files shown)


contracts/src/arbitration/dispute-kits/DisputeKitClassic.sol
```diff
@@ -285,13 +286,11 @@ contract DisputeKitClassic is BaseDisputeKit, IEvidence {
             core.getCurrentPeriod(_coreDisputeID) == KlerosCore.Period.commit,
             "The dispute should be in Commit period."
         );
-        require(_commit != bytes32(0), "Empty commit.");
```
jaybuidl, line 288: We don't really need to remove this to allow changing the commitment. There is no valid case for allowing a 0x00 commitment, it would likely be a bug or a human error so it seems like good UX to have it.

jaybuidl (changes requested):

contracts/src/arbitration/dispute-kits/DisputeKitClassic.sol
donosonaumczuk: Ah sorry I forgot to explain this change.

I think most of users will cast the commit through a UI. So the UI can do the validations. I though would be better to remove that validation from the contracts, as it looks unlikely to happen and now the user can override the vote commitment later in case of submitting a zero one by mistake.

But if you disagree with this reasoning I will push a change reverting that validation removal.

unknownunknown1: I think it's better to keep the require to avoid some flaky behaviour in case someone wants to bypass the UI. Also the user might submit empty commit by mistake without noticing

jaybuidl: Okay let's not change how data validation is done in this particular PR. Data validation could be reviewed separately across the board in a separate PR with a proper discussion of UX vs less code/gas.

donosonaumczuk: Makes sense to treat it separately if needed. I pushed [the change](https://github.com/kleros/kleros-v2/pull/190/commits/233e7d33e3e0f89390a9571cd1e34917b2f21d0f) to keep the validation!

jaybuidl (approved):

PR #201, opened 2022-08-10
Fix: Modify DisputeKitCreated event

alcercu:


alcercu:
This PR is to modify the event DisputeKitCreated (to add the disputeKitID field) and to emit it when a new dispute kit is created (it was missing in the addNewDisputeKit function). A couple of notes:
- The emit DisputeKitEnabled on [line 256](https://github.com/kleros/kleros-v2/blob/41625e2e82cf1a6ac7cd0f595fd181f3f752dbed/contracts/src/arbitration/KlerosCore.sol#L256) and [line 346](https://github.com/kleros/kleros-v2/blob/41625e2e82cf1a6ac7cd0f595fd181f3f752dbed/contracts/src/arbitration/KlerosCore.sol#L346) could be removed if in the subgraph we add all root DisputeKits to the General Court's supported list in the DisputeKitCreated handler. But maybe it's nice to also have the event on-chain?
- Any reason why the _subcourtID is uint256 but the _parent is unit96? This gave me some typing problems when trying to use the variable GENERAL_COURT.

jaybuidl:
> * The emit DisputeKitEnabled on [line 256](https://github.com/kleros/kleros-v2/blob/41625e2e82cf1a6ac7cd0f595fd181f3f752dbed/contracts/src/arbitration/KlerosCore.sol#L256) and [line 346](https://github.com/kleros/kleros-v2/blob/41625e2e82cf1a6ac7cd0f595fd181f3f752dbed/contracts/src/arbitration/KlerosCore.sol#L346) could be removed if in the subgraph we add all root DisputeKits to the General Court's supported list in the DisputeKitCreated handler. But maybe it's nice to also have the event on-chain?

Yes let's keep it on-chain, which doesn't add any extra assumption on the subgraph.

> * Any reason why the _subcourtID is uint256 but the _parent is unit96? This gave me some typing problems when trying to use the variable GENERAL_COURT.

No good reason so it's fixed. It still takes 1 slot of storage but at least typing is more consistent.

jaybuidl (approved):

alcercu:
LGTM 🚀

PR #226, opened 2022-08-25
feat: migration to a L1 blockhash based RNG with significant lookahead as recommended by EIP-4933

### LINKED ISSUE #227: Improve the RNG implementation
The RNG doesn't have to be as good as a VRF, but just unpredictable enough for some serious testing in the medium term.

jaybuidl:
The lookahead time takes into account:
- ~~the post-merge [EIP-4933](https://eips.ethereum.org/EIPS/eip-4399) recommendation of 4 epochs + 4 slots = 132 blocks.~~ <-- does not apply to BLOCKHASH, but to DIFFICULTY only.
- the Arbitrum sequencer finality of 20 blocks.

So the dispute kit obtains a random number based on the hash of a block which is **~~152~~ 20 blocks into the future**.

Also updated the deploy scripts and test cases accordingly.

Resolves #227 

_EDIT: removed the lookahead from EIP-4399, only relevant to DIFFICULTY._

---

From [EIP-4933](https://eips.ethereum.org/EIPS/eip-4399):
image


contracts/src/arbitration/dispute-kits/DisputeKitClassic.sol
```diff
@@ -230,11 +232,13 @@ contract DisputeKitClassic is BaseDisputeKit, IEvidence {
 
         (uint96 subcourtID, , , , ) = core.disputes(_coreDisputeID);
         bytes32 key = bytes32(uint256(subcourtID)); // Get the ID of the tree.
-        uint256 drawnNumber = getRandomNumber();
 
         (uint256 K, uint256 nodesLength, ) = core.getSortitionSumTree(key, 0);
         uint256 treeIndex = 0;
-        uint256 currentDrawnNumber = drawnNumber % core.getSortitionSumTreeNode(key, 0);
+        uint256 currentDrawnNumber = uint256(
+            keccak256(abi.encodePacked(randomNumber, _coreDisputeID, round.votes.length))
```
jaybuidl, line 239: Equivalent of this in v1: https://github.com/kleros/kleros/blob/master/contracts/kleros/KlerosLiquid.sol#L493

jaybuidl, line 239: Equivalent of this in v1: https://github.com/kleros/kleros/blob/master/contracts/kleros/KlerosLiquid.sol#L493

shotaronowhere: So the [difficulty](https://developer.offchainlabs.com/docs/Inside_Arbitrum#more-evm-emulation-details) op code doesn't exist in Arbitrum.

The rng you specified uses the L2 blockhash, so using the lookahead for RANDAO (difficulty op code) isn't relevant. 20 blocks is probably sufficient for a lookahead. 

I just asked on the Arbitrum discord if nitro will include the RANDAO opcode. Maybe we can let them know its a good idea to include, feature request.

shotaronowhere (approved):
This isn't critical, but the RNG_LOOKAHEAD does not need to be so big.

The reference EIP 4399 applies to mainnet for the RANDAO op-code which doesnt exist on arbitrum. 20 blocks should be sufficient.

shotaronowhere (changes requested):
So the [difficulty](https://developer.offchainlabs.com/docs/Inside_Arbitrum#more-evm-emulation-details) op code doesn't exist in Arbitrum.

The rng you specified uses the L2 blockhash, so using the lookahead for RANDAO (difficulty op code) isn't relevant. 20 blocks is probably sufficient for a lookahead.

I just asked on the Arbitrum discord if nitro will include the RANDAO opcode. Maybe we can let them know its a good idea to include, feature request.

jaybuidl:
> So the [difficulty](https://developer.offchainlabs.com/docs/Inside_Arbitrum#more-evm-emulation-details) op code doesn't exist in Arbitrum.
> 
> The rng you specified uses the L2 blockhash, so using the lookahead for RANDAO (difficulty op code) isn't relevant. 20 blocks is probably sufficient for a lookahead.

Right, so looking at the various docs again this is my understanding:
| OPCODE 	| Merge 	| L1                                           	| Arbitrum L2                                       	|
|--------	|-------	|----------------------------------------------	|---------------------------------------------------	|
| 0x44   	| Pre   	| DIFFICULTY<br/> ❌ bad randomness                  	| DIFFICULTY<br/> ❌ constant                   	|
| 0x44   	| Post  	| DIFFICULTY -> PREVRANDAO<br/> ✅ strong randomness 	| DIFFICULTY<br/> ❌ constant                   	|
| 0x40   	| Pre   	| BLOCKHASH<br/> ✅ strong randomness                	| BLOCKHASH<br/> ❌ insecure digest of the chain history 	|
| 0x40   	| Post  	| BLOCKHASH<br/> ❌ bad randomness                   	| BLOCKHASH<br/> ❌ insecure digest of the chain history 	|
| 0x43   	| Both  	| NUMBER<br/>                                       	| NUMBER<br/>⚠️ estimate of L1 for a sequencer chain, a few blocks ahead. 	|

⚠️ An unexpected consequence of this is `blockhash(block.number)` **often returns 0x00** on L2 because block.number is a few blocks ahead, so it hasn't been mined yet on L1

## Relevant info
### [Arbitrum solidity support page](https://developer.offchainlabs.com/docs/Solidity_Support)
`blockhash`
> A value deterministically generated from the L1 block number and state of the inbox

`block.number`
> On a Sequencer chain, block.number returns an “estimate” of the L1 block number at which the Sequencer received the transaction.

### [Arbitrum EVM emulation details](https://developer.offchainlabs.com/docs/Inside_Arbitrum#more-evm-emulation-details):
> `BLOCKHASH` returns a pseudorandom value that is a digest of the chain’s history, but not the same hash value that Ethereum L1 would return at the same block number.

### [Security considerations](https://developer.offchainlabs.com/docs/Security_Considerations):
> Arbitrum's L2 `BLOCKHASH` should not be relied on as a secure source of randomness.

### [How the merge impacts the app layer](https://blog.ethereum.org/2021/11/29/how-the-merge-impacts-app-layer/) and [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399)
> Post-merge, the BLOCKHASH opcode will still be available for use, but given that it will no longer be forged through the proof of work hashing process, the pseudorandomness provided by this opcode will be much weaker.
>
> Relatedly, the DIFFICULTY opcode (0x44) will be updated and renamed to PREVRANDAO. Post-merge, it will return the output of the randomness beacon provided by the beacon chain. This opcode will thus be a stronger, albeit still biasable, source of randomness for application developers to use than BLOCKHASH.
>
> Pre-merge, we see the 0x44 opcode returns the difficulty field in the block header. Post-merge, the opcode, renamed to PREVRANDAO.

jaybuidl:
>The rng you specified uses the L2 blockhash, so using the lookahead for RANDAO (difficulty op code) isn't relevant. 20 blocks is probably sufficient for a lookahead.

Done ✅  @shotaronowhere

contracts/src/arbitration/dispute-kits/DisputeKitClassic.sol
jaybuidl: ✅

shotaronowhere (dismissed):
so I think this is a 'good enough' temporary non-production ready solution.

comment about the RNG_LOOKAHEAD. 20 blocks sort of relates to L2 finality, but in the sense that the inbox determines the L2 state. So as soon as the L1 txn to submit a batch of txns in the inbox finalizes on L1, you know it will be included in arbitrum. 

per the arbitrum docs, the blockhash on L2 is not the blockhash on L1, but determined by the inbox and L1 block number. And the sequencer is centralized to despite any block lookahead, aslong as the sequencer is centralized, it can completely determine the blockhash by selecting the txns in a batch appropriately. So nonetheless, better than an incremental rng or the mock rngs that exist right now.

unknownunknown1 (approved):

PR #233, opened 2022-08-25
refactor: rng inheritance simplified, removed unused incentive mechanism

jaybuidl:


contracts/src/rng/BlockhashRNG.sol
```diff
@@ -1,64 +1,50 @@
...
+        if (randomNumber != 0) {
+            return randomNumber;
         }
 
-        if (randomNumber[_block] != 0) {
-            // If the number is set.
-            uint256 rewardToSend = reward[_block];
-            reward[_block] = 0;
-            payable(msg.sender).send(rewardToSend); // Note that the use of send is on purpose as we don't want to block in case msg.sender has a fallback issue.
+        if (_block < block.number && randomNumbers[_block] == 0) {
```
shotaronowhere, line 38: no need to check randomNumbers[_block] == 0, this is already known from line 34.

contracts/src/rng/BlockhashRNG.sol
```diff
@@ -1,64 +1,50 @@
...
+            // The random number is not already set and can be.
+            if (blockhash(_block) != 0x0) {
+                // Normal case.
+                randomNumbers[_block] = uint256(blockhash(_block));
```
shotaronowhere, line 42: not a big deal, but would be cheaper to do 
`randomNumber = uint256(blockhash(_block));`
`randomNumbers[_block] = randomNumber;`

and remove line 48. Similarly for the else block.

`randomNumber = uint256(blockhash(block.number - 1));`
`randomNumbers[_block] = randomNumber;`

shotaronowhere (changes requested):
the changes are actually pretty low priority. just some small gas optimization, but might as well

contracts/src/rng/IncrementalNG.sol
```diff
@@ -17,19 +17,19 @@ contract IncrementalNG is RNG {
...
 
     /**
-     * @dev Contribute to the reward of a random number. All the ETH will be lost forever.
-     * @param _block Block the random number is linked to.
+     *  @dev Request a random number.
+     *  @param _block Block the random number is linked to.
      */
-    function contribute(uint256 _block) public payable override {
-        /* NOP */
+    function requestRandomness(uint256 _block) external override {
```
unknownunknown1, line 23: This function is not used anywhere? Do we really want to keep it?

unknownunknown1 (approved):

PR #252, opened 2022-09-01
feat(KC): add sortition tree module

unknownunknown1:

PR #383, opened 2022-11-07
feat(KlerosCore): disable subcourt

unknownunknown1:


unknownunknown1:
The simplest solution to this is to mark a specific court as disabled, create a dispute for this court, set the parameters accordingly so it's impossible to stake and vote in this court, and then remove all drawn jurors upon execution

Note that at first we discussed the possibility of using the existing unstaking functionalitty of the court https://github.com/kleros/kleros-v2/blob/d751b6c210b6b05cc2c321a02de5fddb6a00a77f/contracts/src/arbitration/KlerosCore.sol#L821
but we decided that'd be too excessive since it unstakes from the parent courts as well. So the proposed solution is basically the advanced version of it.

The alternative to this would be to manually store every juror staked in the court and if we want to disable subcourt we just iterate through them and unstake. I don't think it's more optimal though

jaybuidl:
Blocked by #252. Will review when it's merged because the staking code is likely affected.

PR #389, opened 2022-11-13
feat(Governor): add contract

unknownunknown1:


contracts/src/arbitration/KlerosGovernor.sol
```diff
@@ -0,0 +1,426 @@
...
+            reservedETH = reservedETH.subCap(sumDeposit);
+        } else {
+            session.status = Status.DisputeCreated;
+            uint256 arbitrationCost = arbitrator.arbitrationCost(arbitratorExtraData);
+            session.disputeID = arbitrator.createDispute{value: arbitrationCost}(
+                session.submittedLists.length,
+                arbitratorExtraData
+            );
+            session.sumDeposit = session.sumDeposit.subCap(arbitrationCost);
+
```
jaybuidl, line 297: missing `sessions.push();` ?

unknownunknown1: When the list is disputed we don't push the session right away. It's one in `rule()` once the dispute is resolved

PR #460, opened 2022-12-16
Fix(contracts): Remove court and subcourt naming inconsistencies

alcercu:


jaybuidl (changes requested):
So far so good. Let's add the corresponding changes to subgraph/ and web/ at the same time.

PR #511, opened 2023-01-22
fix(IMetaevidence): remove evidence

unknownunknown1:

PR #541, opened 2023-01-27
Interfaces v2

jaybuidl:
Supersedes #511


contracts/src/arbitration/IArbitrableV2.sol
```diff
@@ -0,0 +1,44 @@
...
+     * @param _arbitrator The arbitrator of the contract.
+     * @param _disputeID ID of the dispute in the Arbitrator contract.
+     * @param _externalDisputeID Unique identifier from the dispute creator that is linked to this dispute.
+     * @param _disputeContextUri IPFS path to dispute context, example: '/ipfs/Qmarwkf7C9RuzDEJNnarT3WZ7kem5bk8DZAzx78acJjMFH/disputecontext.json'
+     */
+    event Dispute(
+        IArbitrableV2 indexed _arbitrator,
+        uint256 indexed _disputeID,
+        uint256 _externalDisputeID,
+        string _disputeContextUri
```
greenlucid, line 26: In defense of `MetaEvidence` event (or equivalent `DisputeContext`), if you emit this separately, and then you reference it with an ID, that's much cheaper.

What should I assume the arbitrator do in this instance, store the `_disputeContextUri` in storage? Then every challenge will require loading this string from storage (which could be ~6 cold SLOAD or 12k gas when making a challenge).

If you instead were to have the event and you just use an ID, then, that's just 1 cold SLOAD.

PR #543, opened 2023-01-27
Arbitrator ruling event

jaybuidl:
To make indexing more convenient


jaybuidl:
Subgraph and web change not included.

Issue #538, opened 2023-01-27
Upload the policies to Estuary (instead of the Kleros gateway)


jaybuidl:

jaybuidl:
Done manually for the General court only.

PR #561, opened 2023-02-02
Upload the policies to Estuary (instead of the Kleros gateway)

Params10:
- Wrote a script to upload the policies on estuary and send a transaction to set the policies from policyRegistery.


jaybuidl:
Resolves #538

Issue #320, opened 2022-09-30
Web3 Provider for WalletConnect


jaybuidl:
Integrate [WalletConnect's Web3 Modal](https://web3modal.com/)

abhiram11:
Hi!

To my understanding, the proposal is for `<ConnectButton />` to use `wagmi` instead of `@web3-react/core` or `Web3Provider from @ethersproject/providers`, right? I would like to know more about the possible changes to be made so I can make the required contributions.

Thanks : )

Arun9650:
hello, I have integrated the Wagmi.sh to connect to the wallet.

jaybuidl:
Thanks for your interest abhiram11 and arun9650, there was a bit more going into this, it will be covered by the migration to wagmi in PR #670.

PR #670, opened 2023-04-05
Refactor(web): migration to wagmi and viem

alcercu:

PR #680, opened 2023-04-10
Wagmi.sh/close#320

Arun9650:
I have integrated the wagmi.sh for issue #320, but it will only work for meta-mask till now as the maintainer sees the code and approve it I will try to add the rest of the connectors like walletConnect and coin-base-wallet


jaybuidl:
Duplicates #670

Issue #689, opened 2023-04-11
Cases search by disputeId


jaybuidl:
![image](https://user-images.githubusercontent.com/22213980/231136131-d306509b-a6ec-440c-9237-c86c465382ca.png)

Arun9650:
I world like to work on this issue

PR #751, opened 2023-04-18
feat(web): cases search by disputeId

Arun9650:
Hello, I have fixed the issue of https://github.com/kleros/kleros-v2/issues/689 by adding a fetch request every time a user asks for a particular case by entering the case id
I achieve this by creating a new file called useCasesQueryById which will request the case id. After that, I pass the obtained data from the "useCasesQueryById" hook to the CasesDisplay Component to show the case.

https://user-images.githubusercontent.com/76466509/232862035-4a95527e-a43e-4434-9540-e464a9e6e23f.mp4


jaybuidl:
It was implemented separately but thanks for the PR anyway.

PR #823, opened 2023-05-11
fix(web): add MouseEvent & IPolicy types

kemuru:
(some lines got automatically formatted when doing the commit)

added MouseEvent & IPolicy types instead of any types.

There are still a few "any" types but starting to think it would be more confusing to remove them than to let them stay, explanation:

**provider: any**

Screenshot from 2023-05-11 23-31-05

this library should have included the type for this I guess, but its a bit messy and not clear. I found this hack, you could add to the package.json the following:

```
"resolutions": {
    "zustand": "^3.6.9"
  }
```

 but still didnt work properly as it does with "any" and some other error appeared

**maybeObjects: any[] | any**

Screenshot from 2023-05-11 23-33-11

in this function you don't really know what's passing as parameter, so makes sense its any, you could make the type "unknown" too I guess, but its kind of the same

**({ query, variables }: { query: string; variables?: any }**

Screenshot from 2023-05-11 23-35-51

this took a while and I ended up thinking it could maybe be more confusing to add a type for this than to leave it like this. This "variables" parameter seems to be the input parameters for the query. At first I did this

Screenshot from 2023-05-11 21-10-18

but then I realized that you could maybe input objects as the query too, not just 1 string. so, not sure.


alcercu (approved):
lgtm

PR #900, opened 2023-06-12
chore(contracts): uml storage viz

shotaronowhere:
This PR includes a script to produce uml diagrams of the storage layout of any contract. There was a small modification to store a "user defined type" IERC20 interface as an address instead and cast it where necessary (actually nowhere, the type was cast to an address where used)

image

This visualization will be good to make sure we planned the contract upgradability correctly.


contracts/test/storage/KlerosCore-old.dot
jaybuidl, line 1: can we remove this old file?

jaybuidl:
Let's dig in the sol2uml docs and find a solution to the custom interface types (so we don't have to replace every types to an address in the contracts) before merging.

Issue #916, opened 2023-06-14
UI: Links to Help resources and Kleros Solutions


jaybuidl:
This is a follow-up on #879

## Kleros Solutions links
Escrow: https://escrow.kleros.io
Tokens: https://tokens.kleros.io
PoH: https://app.proofofhumanity.id
Curate: https://curate.kleros.io
Resolver: https://resolve.kleros.io
Linguo: https://linguo.kleros.io
Governor: https://governor.kleros.io

## Help links
Onboarding: later link to the page for this feature #266 
Get Help: https://t.me/kleros
Report a Bug: https://github.com/kleros/kleros-v2/issues
DApp Guide: https://docs.kleros.io/products/court-v2
Crypto Beginner's Guide: https://ethereum.org/en/wallets/
FAQ: https://docs.kleros.io/kleros-faq

## Small bug

When switching tab in the browser, it closes the popup. Can we prevent this?

## UI
![image](https://github.com/kleros/kleros-v2/assets/22213980/32b094fe-4993-469d-80a3-ef90a2bb91a5)

PR #961, opened 2023-06-19
fix: fixed links and modal behavior

nhestrompia:
fixed issues mentioned in #916, added links and modified behavior of modals. Now modals don't close when user changes tabs or window.

For onboarding link, waiting this #266


alcercu (approved):
lgtm

PR #965, opened 2023-06-21
fix: update notUndefined usage on some files

nhestrompia:
updated isUndefined function usage on some files that still uses the old version of the function.


alcercu (approved):
lgtm

PR #980, opened 2023-06-26
fix(scripts): handle no mock contract in simulations scripts

kemuru:
error gets triggered when using different chain than localhost in hardhat


alcercu (dismissed):
lgtm

PR #1037, opened 2023-07-13
feat(subgraph): add fee token to subgraph

kemuru:
feat(subgraph): add fee token to subgraph (with @nhestrompia )


subgraph/schema.graphql
```diff
@@ -188,6 +189,14 @@ type Counter @entity {
   casesRuled: BigInt!
 }
 
+type FeeToken @entity {
+  id: ID! # The address of the ERC20 token.
+  accepted: Boolean!
+  rateInEth: BigInt!
+  rateDecimals: Int!
+  rounds: [Round!] @derivedFrom(field: "feeToken")
```
alcercu, line 197: lets add a `totalPaid` and `totalPaidInETH` fields here

subgraph/schema.graphql
alcercu, line 1: on this file let's modify the `TokenAndETHShift` entity:

tokenAmount => pnkAmount
ethAmount => (same name, but now it will be the value of the feeToken shift in ETH)

new fields:
feeToken => the fee token used on this shift (if any)
feeTokenAmount => amount paid using the feeToken

subgraph/src/KlerosCore.ts
alcercu, line 1: on this file, we should use the `FeeToken` entity on the `handleTokenAndETHShift` function and adapt the logic accordingly

alcercu (changes requested):

alcercu (approved):
lgtm

PR #1055, opened 2023-07-26
chore: fix linter error

### LINKED ISSUE #920: Possible bug in the upload script
-> https://sonarcloud.io/project/issues?open=AYfxL_yD7ytXykK8kLuh&id=kleros_kleros-v2

anukul:
Fixes #920 

No bug, warning because the promise constructor was returning another promise (`fetch`), instead of returning `void`.
This can be fixed by just changing
`new Promise((resolve, reject) => fetch(...))` to
`new Promise((resolve, reject) => { fetch(...) })`

but we don't necessarily need another new promise, since `fetch` already returns a promise.
Here, it is semantically preferable to throw an `Error` type, so type annotations are also updated accordingly.


jaybuidl (approved):

jaybuidl:
Thanks @anukul !

PR #1105, opened 2023-08-04
fix: interaction-popups

### LINKED ISSUE #1097: Popup after voting: incorrect when voting without commit/reveal
When there is no commit/reveal, it is misleading to tell the juror that his vote is secret and that he must come back later to reveal.

image

nhestrompia:
Fixes #1097


alcercu (approved):
lgtm

Issue #1103, opened 2023-08-04
Dispute Preview (dev tool)


jaybuidl:
[**Link to the mockup**](https://viewer.diagrams.net/?tags=%7B%7D&highlight=0000ff&edit=_blank&layers=1&nav=1&page-id=lmbbVLH0Eg7425TmWPBL&title=Evidence%20Module%20and%20Dispute%20Template%20Registry.drawio#Uhttps%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1SRvwTg7eyj0yDKzkubOkn7ALeH_ZyeGe%26export%3Ddownload)

![image](https://github.com/kleros/kleros-v2/assets/22213980/b5c4469c-9560-4679-aa58-799e0e295a7e)

jaybuidl:
Reopening: work will continue in a different PR.

https://github.com/kleros/kleros-v2/pull/1165#issuecomment-1906778483

PR #1107, opened 2023-08-04
Feat(web): add dispute template previewer

alcercu:


jaybuidl:
Partially addresses #1103, without the variables and mappings.

PR #1108, opened 2023-08-04
fix: voting-results-and-display

### LINKED ISSUE #1085: Voting history: show the history component during voting period
Rationale: to see the history of previous rounds voting history

Currently it shows only after voting.

nhestrompia:
Fixes #1085 and other edge cases with disputeTemplate


alcercu (changes requested):
I see `getChoiceText` and `getVoteChoice` are quite similar, could we abstract that function?

alcercu (approved):
lgtm

PR #1131, opened 2023-08-09
fix(web): jurorodds overflow fix

kemuru:


alcercu (changes requested):
let's abstract the `bigIntRatioToPercentage` into a hook using `useMemo` and use that hook on the dashboard coherency component too

alcercu (approved):
lgtm

PR #1136, opened 2023-08-09
Feat/upgrade sortition module

zmalatrax:
This PR implements an ERC-1822-compliant (UUPS) & ERC-1967-compliant proxy, and make the SortitionModule upgradeable.

# Proxy & Proxiable
## Contracts

The UUPS Proxy and Proxiable contracts implementation are custom, based on OpenZeppelin, Andrei & Synthetix work, as described [here]([https://gist.github.com/zmalatrax/e3e46044611d57befaf12c9f437b2e0f](https://gist.github.com/zmalatrax/e3e46044611d57befaf12c9f437b2e0f)).

They can be found [here](https://github.com/kleros/kleros-v2/tree/feat/upgrade-sortition-module/contracts/src/proxy). No review has been done yet.
## Testing

In order to test the proxy mechanisms (forwarding, upgrading...), I made different [mock implementations](https://github.com/kleros/kleros-v2/blob/feat/upgrade-sortition-module/contracts/src/proxy/mock/MockImplementations.sol), inspired by OpenZeppelin's tests.

The tests can be found [here](https://github.com/kleros/kleros-v2/blob/feat/upgrade-sortition-module/contracts/test/proxy/index.ts).  
# SortitionModule

## Transition to Upgradeable
Making a contract upgradeable requires a few steps:
- Inherit from `UUPSProxiable`
- Move the constructor to an `initialize` function
	- Add an `bool private initialized` storage variable, to allow initializing only once.
	- For security purposes, initializing the implementation when deployed (in the constructor) is needed to avoid someone taking over the implementation and potentially bricking the proxy (UUPS-related issue)
- Override the `_authorizeUpgrade(address)` function to define access control for upgradeability.
- Add an arbitrary gap (`uint256[50] __gap`)  if this contract is to be inherited from (a Parent contract), to allow updating the storage layout without causing storage collision with Child storage. Further explained [here](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#storage-gaps)
  It is actually not needed for the SortitionModule.
## Testing

I haven't written specific tests for the upgradeability mechanism of the newly upgradeable SortitionModule but it passes the already made test suite. To run tests, the deployment script had to be modified to deploy the proxy and the implementation, thanks to hardhat-deploy. 

The [DisputeKitClassic test](https://github.com/kleros/kleros-v2/commit/eccad1a45a762fd494f87ccac53a386bb66608b4#diff-b717549f72be055210da965a49674276de00dfbd2aa5ac96a1bff8875c2b419a) had to be modified, as it deploys a SortitionModule in it.
# Deployment Script

The SortitionModule is deployed by [`00-home-chain-arbitration.ts`](https://github.com/kleros/kleros-v2/blob/feat/upgrade-sortition-module/contracts/deploy/00-home-chain-arbitration.ts).

The UUPS proxy deployment requires 2 transactions:
1. Deploying the implementation contract
2. Deploying the proxy contract (and initializing it)

As the `KlerosCoreAddress` is computed in advance, based on the nonce to be set for the SortitionModule, the used nonce will be `+2` rather than the previous `+1`.
## hardhat-deploy

Deploying a custom UUPS proxy is not well documented in hardhat-deploy, here is the process:

```typescript
const sortitionModuleDeployment = await deploy("SortitionModule", {
    from: deployer,
    proxy: {
      proxyContract: "UUPSProxy",
      proxyArgs: ["{implementation}", "{data}"],
      execute: {
        init: {
          methodName: "initialize",
          args: [deployer, KlerosCoreAddress, 1800, 1800, rng.address, RNG_LOOKAHEAD], // minStakingTime, maxFreezingTime
        },
      },
    },
    log: true,
    args: [],
  });
```

- The first argument of `deploy` is the deployment name. If it matches a contract name, it will deploy this contract, but it is not necessary to match a contract. If the deployment name doesn't match a contract, the property `contractName` must be defined with the implementation to be deployed in the second argument object. It is necessary if you want to deploy multiple instances of the same contract (e.g in [DisputeKitClassic](https://github.com/kleros/kleros-v2/commit/eccad1a45a762fd494f87ccac53a386bb66608b4#diff-b717549f72be055210da965a49674276de00dfbd2aa5ac96a1bff8875c2b419a))

- The `proxy` object enables the proxy deployment instead of a direct deployment.

- `proxyContract`: The proxy contract name.
- `proxyArgs`: The arguments template of the proxy constructor.  If not specified, it relies on a [default value](https://github.com/wighawag/hardhat-deploy/blob/1a216e1584fd10262b365fa0469d099cd094ead4/src/helpers.ts#L1178) `proxyArgsTemplate = ["{implementation}", '{admin}', "{data}"]` which doesn't suit UUPS Proxies.
  The only UUPS proxy natively supported by hardhat-deploy is a version of the OpenZeppelin's one, the `proxyArgsTemplate` is updated along other parameters. See this [snippet](https://github.com/wighawag/hardhat-deploy/blob/1a216e1584fd10262b365fa0469d099cd094ead4/src/helpers.ts#L1259-L1264).
  - `execute` object with two properties, `init` & `onUpgrade` to define methods to call respectively on first deployment and other deployments of the proxy.
    These two objects, `init` & `onUpgrade` have the same structure:
    - `methodName`: name of the function to be called
    - `args`: Array containing the different arguments used when calling `methodName`.

`onUpgrade` hasn't been defined for the SortitionModule.
## Known limitations

# Misc

### DIFF
--- a/contracts/src/proxy/UUPSProxiable.sol
+++ b/contracts/src/proxy/UUPSProxiable.sol
@@ -1,0 +1,134 @@
+//SPDX-License-Identifier: MIT
+
+/**
+ *  @authors: [@malatrax]
+ *  @reviewers: []
+ *  @auditors: []
+ *  @bounties: []
+ *  @deployments: []
+ */
+pragma solidity 0.8.18;
+
+/**
+ * @title UUPS Proxiable
+ * @author Simon Malatrait <simon.malatrait@grenoble-inp.org>
+ * @dev This contract implements an upgradeability mechanism designed for UUPS proxies.
+ * The functions included here can perform an upgrade of an UUPS Proxy, when this contract is set as the implementation behind such a proxy.
+ *
+ * IMPORTANT: A UUPS proxy requires its upgradeability functions to be in the implementation as opposed to the transparent proxy.
+ * This means that if the proxy is upgraded to an implementation that does not support this interface, it will no longer be upgradeable.
+ *
+ * A security mechanism ensures that an upgrade does not turn off upgradeability accidentally, although this risk is
+ * reinstated if the upgrade retains upgradeability but removes the security mechanism, e.g. by replacing
+ * `UUPSProxiable` with a custom implementation of upgrades.
+ *
+ * The `_authorizeUpgrade` function must be overridden to include access restriction to the upgrade mechanism.
+ */
+abstract contract UUPSProxiable {
+    // ************************************* //
+    // *             Event                 * //
+    // ************************************* //
+
+    /**
+     * Emitted when the `implementation` has been successfully upgraded.
+     * @param newImplementation Address of the new implementation the proxy is now forwarding calls to.
+     */
+    event Upgraded(address indexed newImplementation);
+
+    // ************************************* //
+    // *             Storage               * //
+    // ************************************* //
+
+    /**
+     * @dev Storage slot with the address of the current implementation.
+     * This is the keccak-256 hash of "eip1967.proxy.implementation" subtracted by 1, and is
+     * validated in the constructor.
+     * NOTE: bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1)
+     */
+    bytes32 private constant IMPLEMENTATION_SLOT = 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
+
+    /**
+     * @dev Storage variable of the proxiable contract address.
+     * It is used to check whether or not the current call is from the proxy.
+     */
+    address private immutable __self = address(this);
+
+    // ************************************* //
+    // *             Governance            * //
+    // ************************************* //
+
+    /**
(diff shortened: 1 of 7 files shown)


contracts/src/proxy/UUPSProxy.sol
```diff
@@ -0,0 +1,108 @@
+//SPDX-License-Identifier: MIT
+
```
jaybuidl, line 2: Add a comment "Adapted from <github url to the OZ contract"

contracts/src/proxy/UUPSProxiable.sol
```diff
@@ -0,0 +1,134 @@
+//SPDX-License-Identifier: MIT
+
```
jaybuidl, line 2: Add a comment "Adapted from <github url to the OZ contract"

PR #1158, opened 2023-08-15
feat(web): drawing info message in Voting tab, times up in period tim…

kemuru:
drawing info message in voting tab, and also changed 0h 0m 0s for this:

image


alcercu (dismissed):
lgtm

PR #1184, opened 2023-08-23
feat(web): disable appeal tab on last round and add info message

### LINKED ISSUE #1183: Web: let the user know the dispute is on last round...
...and appeal is not possible anymore. Instead of showing these large numbers on the Appeal tab, we should think of something else

IMG_1910

kemuru:
closes #1183


alcercu (approved):
lgtm

PR #1186, opened 2023-08-25
fix(web): price-fetching

nhestrompia:
fixes price fetching error due to dynamic contract addresses


jaybuidl (changes requested):
It would be better to make new functions explicitly referencing mainnet, like `usePNKAddressMainnet()`

jaybuidl (approved):

PR #1211, opened 2023-09-02
feat(web): desktop responsiveness navbar, footer, community section, dashboard juror info, home page..

kemuru:
this PR contains Desktop responsiveness for: Navbar, Footer, Community Section, Dashboard's Juror Info, Home page, etc


alcercu (commented):
I would like to maintain the mobile first philosophy, as I mentioned on the `smallScreenStyle` comment, the changes would require to take what's inside the `smallScreenStyle` function, move it outside, and then use a `landscapeStyle` to style for bigger screens.
I know that a lot of work has gone into this, so I won't push this too much, if your call is to go with this PR @kemuru @nhestrompia @jaybuidl I'm okay with it

alcercu:
> > I would like to maintain the mobile first philosophy, as I mentioned on the `smallScreenStyle` comment, the changes would require to take what's inside the `smallScreenStyle` function, move it outside, and then use a `landscapeStyle` to style for bigger screens. I know that a lot of work has gone into this, so I won't push this too much, if your call is to go with this PR @kemuru @nhestrompia @jaybuidl I'm okay with it
> 
> yeah I think that makes sense too. I can invert the queries today maybe. Which landscapeStyle breakpoint should we use?

Depends on what we want for tablet 🤔 If we want landscape tablet to be like desktop, I would go with 900px. If we want it to look like mobile, I would go with 1200px. (I incline towards 900px, so tablet portrait would look like mobile and tablet landscape would look like desktop)

alcercu (approved):
lgtm

PR #1219, opened 2023-09-06
feat(web): cases-grid-and-list-display

### LINKED ISSUE #688: Cases list view (Desktop)
image

nhestrompia:
Closes #688 , also fixes Dashboard page cases section


jaybuidl (changes requested):
I found a few issues:
- In mobile format, the list view exceeds the screen width. Either we try to support horizontal scrolling (probably too much work), either we disable the list view button until the screen is wide enough.
- There is an intermediate width (at 00:12s in the video) where the last columns headers wrap to the next line
- In desktop view, some disputes info is not aligned (court, reward, deadline) and the category is missing when the title is somewhat long (at 00:25s in the video)


https://github.com/kleros/kleros-v2/assets/22213980/d0a435a0-1883-4267-8c22-100b626d76f9

alcercu (changes requested):
- There is horizontal scrolling on mobile on the cases page, one approach to fixing this would be to fix the width to 312px on mobile, and 380px on landscape.
- When using the list view, if the title of the dispute is too big, it looks bad, we either decide to show only two lines and then "..." or we allow them to expand vertically 🤔 I am inclined towards the first option.
- For the grid <> list switch, if we use grey it looks like it's disabled 🤔 I would either show both in blue with the selected one with some kind of highlight, like glow, or show only the one you can change towards (i.e. if we are on grid view, only show the icon to change to list view, and viceversa)

PR #1225, opened 2023-09-11
Reduction of KlerosCore bytecode size

jaybuidl:


jaybuidl:
Some of these changes have been cherry-picked and added into #1214

PR #1282, opened 2023-10-18
Subgraph notification

shotaronowhere:
This PR adds features to the subgraph to support polling notification systems.

- ipfs feature fetches court information (name, description, required skills)
- voteID key in the Round entity is renamed to voteIDNum to avoid a collision with the [graphql-codegen](https://www.npmjs.com/package/@graphql-codegen/typescript) package which generates types for graph queries.

---

## PR-Codex overview
This PR focuses on adding new functionality related to court jumps and period index counters. 

### Detailed summary
- Added `user.rounds` array to `User` entity in `User.ts`
- Removed `classicDispute.jumped` property in `ClassicDispute.ts`
- Added `handleCourtJump` event handler in `subgraph.yaml`
- Added `getAndIncrementPeriodCounter` function in `PeriodIndexCounter.ts`
- Updated `useClassicAppealQuery` and `useDrawQuery` hooks in `useClassicAppealQuery.ts` and `useDrawQuery.ts`
- Added `getLocalRounds` function in `getLocalRounds.ts`
- Updated code in `DisputeCard/index.tsx`, `useClassicAppealContext.tsx`, `Overview.tsx`, `Voting/index.tsx`, `CaseDetails/VotingHistory.tsx`, and `KlerosCore.ts`
- Added `handleCourtJump` event handler in `KlerosCore.ts`

> The following files were skipped due to too many changes: `subgraph/src/KlerosCore.ts`, `web/src/components/Verdict/DisputeTimeline.tsx`, `subgraph/schema.graphql`


subgraph/schema.graphql
```diff
@@ -142,16 +146,18 @@ type Court @entity {
...
   period: Period!
   ruled: Boolean!
   currentRuling: BigInt!
   tied: Boolean!
   overridden: Boolean!
-  lastPeriodChange: BigInt!
-  lastPeriodChangeBlockNumber: BigInt!
   periodDeadline: BigInt!
+  periodNotificationIndex: BigInt!
+  lastPeriodChangeTs: BigInt!
```
alcercu, line 159: Could we go without the "Ts" suffix on `lastPeriodChange` ? Mostly to keep the same name as the one in the contract.

subgraph/schema.graphql
```diff
@@ -179,8 +191,9 @@ type Draw @entity(immutable: true) {
   dispute: Dispute!
   round: Round!
   juror: User!
-  voteID: BigInt!
+  voteIDNum: BigInt!
```
alcercu, line 194: Same here, I don't think the "Num" suffix is needed.

subgraph/src/KlerosCore.ts
```diff
@@ -126,10 +126,21 @@ export function handleNewPeriod(event: NewPeriod): void {
   }
 
   dispute.period = newPeriod;
-  dispute.lastPeriodChange = event.block.timestamp;
+  dispute.lastPeriodChangeTs = event.block.timestamp;
+  dispute.lastPeriodChangeBlockNumber = event.block.number;
+  let counter = PeriodIndexCounter.load(newPeriod);
```
alcercu, line 131: I think you can use `getAndIncrementCounter` here instead.

subgraph/src/KlerosCore.ts
```diff
@@ -126,10 +126,21 @@ export function handleNewPeriod(event: NewPeriod): void {
...
+  if (!counter) {
+    counter = new PeriodIndexCounter(newPeriod);
+    counter.counter = BigInt.fromI32(0);
+  }
+  dispute.periodNotificationIndex = counter.counter;
+  counter.counter = counter.counter.plus(ONE);
+  counter.save();
   dispute.lastPeriodChangeBlockNumber = event.block.number;
```
alcercu, line 139: This line is duplicated with 130.

subgraph/src/KlerosCore.ts
```diff
@@ -126,10 +126,21 @@ export function handleNewPeriod(event: NewPeriod): void {
...
   if (newPeriod !== "execution") {
-    dispute.periodDeadline = event.block.timestamp.plus(court.timesPerPeriod[event.params._period]);
+    const contract = KlerosCore.bind(event.address);
```
alcercu, line 141: If possible, I would prefer to keep the use of the subgraph's `timesPerPeriod` from the court entity, rather than binding the contract and calling it. The second method is way slower and hurts sync performance.

subgraph/src/PolicyRegistry.ts
```diff
@@ -1,5 +1,6 @@
 import { PolicyUpdate } from "../generated/PolicyRegistry/PolicyRegistry";
 import { Court } from "../generated/schema";
+import { ipfs, json, Bytes, TypedMap, JSONValue } from "@graphprotocol/graph-ts";
```
alcercu, line 3: I would prefer if we could do without the `ipfs` query as it's not clear if it will be supported by the decentralized network.

subgraph/src/entities/PeriodIndexCounter.ts
```diff
@@ -0,0 +1,14 @@
+import { PeriodIndexCounter } from "../../generated/schema";
+import { BigInt } from "@graphprotocol/graph-ts";
+
+export function getAndIncrementCounter(id: string): BigInt {
```
alcercu, line 4: Could we change the function name to `getAndIncrementPeriodCounter` or `getAndIncrementPeriodIndexCounter` mostly to avoid confusion with the `Counter` entity.

alcercu (changes requested):

subgraph/schema.graphql
shotaronowhere: So there is a collision in namespace when the graphql-codegen package generates types for graph queries.

There are Vote entities if an ID field [L31](https://github.com/kleros/kleros-v2/blob/42dfcd21c19e7256d9ed0d10ce8f4d298e0f6a5f/subgraph/schema.graphql#L31C10-L31C10). The generated types through errors, and more generally, its probably better practice to disambiguate the voteID here since it is an argument of an event, and not the id key of a vote entity.

I'm not sure a good alternative. . . . 

eventArgVoteID

subgraph/src/KlerosCore.ts
shotaronowhere: perhaps this cuts corners. . .  but the real problem being avoided here is that the appealDecision handler doesn't detect subcourt jumps so reading the contract avoids depending on the graph schema. We can fix this properly though. . .

subgraph/src/entities/PeriodIndexCounter.ts
shotaronowhere: I've updated this to getAndIncrementPeriodCounter

subgraph/src/KlerosCore.ts
shotaronowhere: I added a court jump handler and reverted by changes to lookup the court entity.

subgraph/src/KlerosCore.ts
shotaronowhere: fixed

subgraph/src/KlerosCore.ts
shotaronowhere: fixed

subgraph/schema.graphql
shotaronowhere: done

subgraph/src/PolicyRegistry.ts
shotaronowhere: removed

shotaronowhere:
note, I removed jump from the classicDisputeKit entity.

We can put it back, but the jump property was never properly updated. 

I tried to properly update the jump property, but the potential solutions seem complicated. In fact, the jump property should be abstracted into the DisputeKitDispute interface.

From the perspective of the Dispute entity (core dispute), it tracks DisputeKitDisputes, but, since each ClassicDispute extends the DisputeKitDispute interface, we cannot load a generic DisputeKitDispute entity from a core dispute id. A solution would be to use a switch case loader utility function to handle each entity type for new dispute kits. This appeared to be too much complexity so I just removed the field "jump" for now. If the frontend is relying on this parameter, we can put it back, but it was hardcoded as "false".

alcercu (approved):

jaybuidl (approved):

PR #1331, opened 2023-11-15
Staking rewards: contract, claim pop-up, styled dashboard jurorinfo

kemuru:
frontend work is done for now:

WIP: we need to deploy the Staking Rewards contract, then update the subgraph, then create new queries in the frontend to get and display the correct variables both in the Staking Rewards component and in the claimed popup.


jaybuidl:
Superseded by #1913

PR #1345, opened 2023-12-01
fix(subgraph): casting-fixes

Harman-singh-waraich:
1. Kleros core
    - appears when` Execute` function is called and `TokenAndEthShift` event is fired.
    - stack :-` handleTokenAndETHShift -> updatePenalty -> entities/Penalty.ts/line-9 `, jurorAddress not being casted to 
    HexString, fixed that

2. Dispute Template
    - _templateTag is a indexed string , so it is being hashed(keccak256)
    - when handling in handleDisputeTemplate , toString is being used instead of toHexString(since it's a hash) to convert 
    _templateTag

Cross checked the weird characters for both the subgraphs and confirmed the above.

For cross checking, these are the event logs https://goerli.arbiscan.io/tx/0x0f2b2c6a2b145f5e8889a262db0a3fb2ebb72566f924746d6a3819ff6e2380ac#eventlog
that were triggering the code and resulting in warnings

---

### Detailed summary
- The focus of this PR is to update the code to use the `toHexString()` method when assigning values to `jurorAddress` and `disputeTemplate.templateTag` variables.
- This PR also updates the `disputeTemplate.templateDataMappings` variable to be assigned as a string value.


alcercu (approved):
lgtm

PR #1350, opened 2023-12-06
refactor(web): extract-calc-to-ResponsiveSize-func

Harman-singh-waraich:
Extracted calc function used in css for making properties responsive to a seperate "ResponsiveSize" function

---

### Detailed summary
- Import of `responsiveSize` from `styles/responsiveSize` added to multiple files.
- Usage of `responsiveSize` function to set margin, padding, width, and height values in styled components.
- The `responsiveSize` function calculates responsive values based on the viewport width.

> The following files were skipped due to too many changes: `web/src/components/Verdict/DisputeTimeline.tsx`, `web/src/components/DisputeCard/DisputeInfo.tsx`, `web/src/layout/Header/navbar/Menu/Settings/index.tsx`, `web/src/pages/Home/TopJurors/JurorCard/DesktopCard.tsx`, `web/src/pages/Courts/index.tsx`, `web/src/pages/Cases/index.tsx`, `web/src/pages/Cases/CaseDetails/Voting/index.tsx`, `web/src/components/CasesDisplay/index.tsx`, `web/src/pages/Home/index.tsx`, `web/src/components/Popup/MiniGuides/Onboarding/PnkLogoAndTitle.tsx`, `web/src/components/Popup/ExtraInfo/StakeWithdrawExtraInfo.tsx`, `web/src/pages/Home/CourtOverview/Stats.tsx`, `web/src/pages/Home/TopJurors/Header/DesktopHeader.tsx`, `web/src/styles/responsiveSize.ts`, `web/src/pages/Courts/CourtDetails/index.tsx`, `web/src/layout/Header/DesktopHeader.tsx`, `web/src/pages/Cases/CaseDetails/Overview/Policies.tsx`, `web/src/pages/Cases/CaseDetails/Timeline.tsx`, `web/src/pages/Dashboard/index.tsx`, `web/src/layout/Header/navbar/DappList.tsx`, `web/src/components/DisputeCard/index.tsx`, `web/src/components/EvidenceCard.tsx`, `web/src/components/Popup/Description/Appeal.tsx`, `web/src/components/CasesDisplay/CasesListHeader.tsx`, `web/src/components/Popup/Description/StakeWithdraw.tsx`, `web/src/components/Popup/index.tsx`, `web/src/components/Popup/MiniGuides/MainStructureTemplate.tsx`

### DIFF
--- a/web/src/styles/responsiveSize.ts
+++ b/web/src/styles/responsiveSize.ts
@@ -1,29 +1,12 @@
-import { css, CSSProperties } from "styled-components";
-
-//supported font types
-type FontType = "px" | "rem" | "em" | "vw" | "vh" | "%" | "pt";
-
 /**
- * @description this func applies repsonsivenexx to a css property, the value will range from minSize to maxSize
- * @param property the css property to apply responsive sizes too
+ * @description this func applies repsonsiveness to a css property, the value will range from minSize to maxSize
  * @param minSize the minimum value of the property
  * @param maxSize max value of the property
  * @param minScreen the min screen width at which the property will be at minSize
  * @param maxScreen the max screen width at which the property will be at maxSize
- * @param fontType
- * @returns
  *
  */
-export const responsiveSize = (
-  property: keyof CSSProperties,
-  minSize: number,
-  maxSize: number,
-  minScreen: number = 375,
-  maxScreen: number = 1250,
-  fontType: FontType = "px"
-) =>
-  css({
-    [property]: `calc(${minSize}${fontType} + (${maxSize} - ${minSize}) * (min(max(100vw, ${minScreen}${fontType}), ${maxScreen}${fontType}) - ${minScreen}${fontType}) / (${
-      maxScreen - minScreen
-    }))`,
-  });
+export const responsiveSize = (minSize: number, maxSize: number, minScreen: number = 375, maxScreen: number = 1250) =>
+  `calc(${minSize}px + (${maxSize} - ${minSize}) * (min(max(100vw, ${minScreen}px), ${maxScreen}px) - ${minScreen}px) / (${
+    maxScreen - minScreen
+  }))`;
--- a/web/src/styles/responsiveSize.ts
+++ b/web/src/styles/responsiveSize.ts
@@ -1,0 +1,29 @@
+import { css, CSSProperties } from "styled-components";
+
+//supported font types
+type FontType = "px" | "rem" | "em" | "vw" | "vh" | "%" | "pt";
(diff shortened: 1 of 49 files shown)


web/src/styles/responsiveSize.ts
```diff
@@ -0,0 +1,29 @@
...
+  minSize: number,
+  maxSize: number,
+  minScreen: number = 375,
+  maxScreen: number = 1250,
+  fontType: FontType = "px"
+) =>
+  css({
+    [property]: `calc(${minSize}${fontType} + (${maxSize} - ${minSize}) * (min(max(100vw, ${minScreen}${fontType}), ${maxScreen}${fontType}) - ${minScreen}${fontType}) / (${
+      maxScreen - minScreen
+    }))`,
```
alcercu, line 28: I think the `fontType` at the second part of the formula break it 🤔 Imagine you pass `${responsiveSize("gap", 15, 40, 300, 1250, "vw")}` like in `CasesListHeader`. Now that last part is `min(max(100vw, 300vw), 1250vw) - 300vw) / 950`, but I think this should always be pixels 🤔 Also, @kemuru implemented this function here https://github.com/kleros/escrow-v2/pull/3/files#diff-d510e52f83ee8d12ef666d39227fc45e58234891832fd94dff2a9fcdcc445fcb I think it would be ideal to have the same implementation on both websites.

alcercu (changes requested):

web/src/styles/responsiveSize.ts
Harman-singh-waraich: 1. This site better explains the calc func 

    https://css-tricks.com/snippets/css/fluid-typography/
    https://www.madebymike.com.au/writing/fluid-type-calc-examples/

    so the the function is independent of length unit given that all properties use same unit,
    but in our case like u pointed out in the casesListHeader  we are using mixed units (vw for unit and px for screen 
    size),
    so yea that might break.

    I'll take a look onto that and think i'll remove the fontType from function for now, since we are using px 
    everywhere except two places.

2. also for @kemuru 's implementation of calcMinMax,

      i think his is more simpler, we have to just take into account the screen sizes too ,like in mine implementation.
     And keeping the property out of func is good too

kemuru: @Harman-singh-waraich  yea I think the function should only handle the pixels. but I think your function name is better (more clear than `calcMinMax`) so I would use my function, but changing the name to yours (`responsiveSize`).

alcercu (approved):
lgtm

Issue #1415, opened 2024-01-15
Web, subgraph?: PNK Locked showing as 75 PNK even though the cases are Closed and executed.


kemuru:
This complaint comes from a community user that he has 2 closed cases and has 75 Locked PNK. when it should be 0 PNK Locked since the cases are closed and executed. User address is `0x4Ca2a670bd5505B870B21331a53D9bB4AE6e2084` 

![image (5)](https://github.com/kleros/kleros-v2/assets/102478601/97bad001-8f8f-4c15-afa3-6fe39e6df76a)

PR #1421, opened 2024-01-17
fix(web): incorrect-locked-pnk-display-in-dashboard

Harman-singh-waraich:
in-directly related to issue #1415 

- Issue : Locked PNK in `Dashboard` is incorrect

- Reproduce : stake in a sub court, it will not reflect in `Dashboard`, until a stake is made in General Court
- Fix: switched the data source of locked PNK to `getJurorBalance` on `SortitionModule`, as that is more reliable and gives global lockedPNK


kemuru (dismissed):
lgtm. only thing that concerns me is that we are showing too much decimals, 0 decimals would be more stylish, but on the other hand it's informational. so I think it's ok for now.
image

alcercu (approved):
lgtm

alcercu (changes requested):
I think it's better if we go with only 2 decimals

jaybuidl (approved):

alcercu (commented):
lgtm

PR #1438, opened 2024-01-22
feat(web): add-enhanced-number-display-component

### LINKED ISSUE #1427: Web: dynamic decimals
Don't show unnecessary decimals when last decimals end in "0", that is, they do not provide extra information. So, show decimals by default (up to a certain number of decimals, for example, 2) but if the last decimals are all zeros, the decimals do not appear because they do not provide extra information.

(note, in the second example in reality I have ETH rewards, its something like 0.00002 ETH, but since it's capped to 3 decimals its showing 0, so maybe fix that too).

Screenshot 2024-01-19 at 10 50 15
Screenshot 2024-01-19 at 10 51 51

Harman-singh-waraich:
closes #1427 

- Adds `NumberDisplay` component, that displays the number up to  specified decimals, the precise value is shown in a tooltip when hovered.
- Added in places where user might want to see precise values, (Dashboard, Top Jurors Rewards Display)
Screenshot 2024-01-22 at 18 15 41
Screenshot 2024-01-22 at 18 16 15

---

### Detailed summary:

- Added a new component `NumberDisplay` for displaying numbers with optional units and tooltips.
- Updated multiple files to use the `NumberDisplay` component instead of directly displaying numbers.
- Replaced the display of stake and rewards values with `NumberDisplay` in the `CourtCard`, `Rewards`, `Header`, and `TokenRewards` components.
- Modified the `getAmount` and `getValue` functions in `jurorRewardConfig.ts` to return formatted numbers using `formatEther` and `formatUnits` functions.
- Updated the `getFormattedRewards` function in `jurorRewardConfig.ts` to use the correct coin price based on the token.
- Imported `Tooltip` component from `@kleros/ui-components-library` in the `NumberDisplay` component.

### DIFF
--- a/subgraph/package.json
+++ b/subgraph/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kleros/kleros-v2-subgraph",
-  "version": "0.3.4",
+  "version": "0.3.5",
   "license": "MIT",
   "scripts": {
     "update:core:arbitrum-sepolia-devnet": "./scripts/update.sh arbitrumSepoliaDevnet arbitrum-sepolia core/subgraph.yaml",
--- a/web/src/components/NumberDisplay.tsx
+++ b/web/src/components/NumberDisplay.tsx
@@ -20,9 +20,8 @@ const NumberDisplay: React.FC<INumberDisplay> = ({
   const parsedValue = Number(value);
   const formattedValue = parsedValue % 1 !== 0 ? parsedValue.toFixed(decimals) : parsedValue.toFixed(0);
   const tooltipValue = isCurrency ? `${unit} ${value}` : `${value} ${unit}`;
-  const displayValue = isCurrency
-    ? `${showUnitInDisplay ? unit : ""} ${formattedValue}`
-    : `${formattedValue} ${showUnitInDisplay ? unit : ""}`;
+  const displayUnit = showUnitInDisplay ? unit : "";
+  const displayValue = isCurrency ? `${displayUnit} ${formattedValue}` : `${formattedValue} ${displayUnit}`;
   return (
     <Tooltip small text={tooltipValue} place={place}>
       {displayValue}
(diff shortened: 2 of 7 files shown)


alcercu (changes requested):
looks good, but I think it would look even better if we keep the units in the tooltip too 🤔 like `0.08 ETH` then you hover and you see `0.078 ETH` on the hover tooltip too 🤔

kemuru (changes requested):
when I hover on the ETH reward, I cannot see my actual ETH reward (which is 0.000015 ETH, or 0.00003) its showing only up to 3 decimals. 

maybe the hovered value should not be fixed to certain decimals? just let that be as big as it wants to be?

image

image

kemuru:
> looks good, but I think it would look even better if we keep the units in the tooltip too 🤔 like `0.08 ETH` then you hover and you see `0.078 ETH` on the hover tooltip too 🤔 wdyt?

yes i agree with this also

Harman-singh-waraich:
> when I hover on the ETH reward, I cannot see my actual ETH reward (which is 0.000015 ETH, or 0.00003) its showing only up to 3 decimals.
> 
> maybe the hovered value should not be fixed to certain decimals? just let that be as big as it wants to be?
> 
> image image

Yes, Just noticed the rewards formatting function has set decimals to 3 , fixing that

Harman-singh-waraich:
> looks good, but I think it would look even better if we keep the units in the tooltip too 🤔 like `0.08 ETH` then you hover and you see `0.078 ETH` on the hover tooltip too 🤔 wdyt?

- Added `Unit` to `NumberDisplay`, also added a flag to not show unit while displaying formatted value(the one that shows normally), as in some place (Top Jurors) we don't show unit
- Also fixed coin value calculation in `Juror Rewards`

alcercu (approved):
lgtm

kemuru (approved):
lgtm

PR #1463, opened 2024-01-29
Feat(web)/case cards indicators

Harman-singh-waraich:
- Adds indicators on Dispute card that provide more insights to Jurors
- Adds a `rewardAmount` entity to `ClassicContribution` 
- Fixes a bug in `ensureClassicContributionFromEvent` where it would increase amount even on `Withdraw` event, the amount in withdraw event is the reward/ reimburse amount , not the contributed amount


alcercu (approved):
lgtm

PR #1488, opened 2024-02-07
fix(web): account-for-rta-option-while-brute-forcing

Harman-singh-waraich:
Fixes `Justify&Reveal` vote failing

Screenshot 2024-02-06 at 22 08 00

### Source of error :
When voting, option 0 is for `Refuse to Arbitrate`, so the length of answers is (`i + 1`).
But when brute forcing this was not being taken into account, and the length of answers array was `i` only.
So if someone voted the last option (`i + 1`), the `getSaltAndChoice` function would return `-1 `, thus leading to the error.

---

### Detailed summary

- Updated the `disabled` prop of the `Button` component to include an additional condition `isUndefined(disputeDetails)`
- Added a new answer option with the title "Refuse To Arbitrate" and description "Refuse To Arbitrate" to the `answers` array.


alcercu (approved):
lgtm

PR #1505, opened 2024-02-16
refactor(web): update-case-list-view

Harman-singh-waraich:
- updates the list view in. cases grid

### DIFF
--- a/web/src/components/CasesDisplay/CasesGrid.tsx
+++ b/web/src/components/CasesDisplay/CasesGrid.tsx
@@ -9,7 +9,6 @@ import { decodeURIFilter } from "utils/uri";
 import { DisputeDetailsFragment } from "queries/useCasesQuery";
 import useIsDesktop from "hooks/useIsDesktop";
 import DisputeCard from "components/DisputeCard";
-import CasesListHeader from "./CasesListHeader";
 
 const GridContainer = styled.div`
   display: flex;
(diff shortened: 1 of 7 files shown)


web/src/components/CasesDisplay/CasesGrid.tsx
```diff
@@ -9,7 +9,6 @@ import { decodeURIFilter } from "utils/uri";
 import { DisputeDetailsFragment } from "queries/useCasesQuery";
 import useIsDesktop from "hooks/useIsDesktop";
 import DisputeCard from "components/DisputeCard";
-import CasesListHeader from "./CasesListHeader";
```
alcercu, line 12: If this file is no longer used we could remove it.

web/src/components/DisputeCard/index.tsx
```diff
@@ -67,7 +67,7 @@ const ListTitle = styled.div`
   height: 100%;
   justify-content: start;
   align-items: center;
-  width: ${responsiveSize(240, 300, 900)};
+  width: ${responsiveSize(240, 420, 900)};
```
alcercu, line 70: I get small cards with a lot of wasted space when the display is thinner 🤔 we could go with a grid based solution, similar to what we are doing on the Escrow: https://github.com/kleros/escrow-v2/blob/master/web/src/components/TransactionsDisplay/TransactionsGrid.tsx#L15
image

alcercu (changes requested):
I think it would look better if we give more structure to these two sections.
The one on the left to make sure the case numbers are aligned. And for the section on the right to prevent unwanted wrapping when the category or subcourt name is too long.
image

PR #1568, opened 2024-04-23
Refactor(web): migrate to vite

alcercu:


jaybuidl:
Update from @alcercu:
We need to rework the svgs to make sure they all show properly.

PR #1617, opened 2024-06-13
feat(web): Evidence: display submission timestamp

ShubhamParkhi:


web/src/components/EvidenceCard.tsx
```diff
@@ -96,6 +97,10 @@ const DesktopText = styled.span`
   )}
 `;
 
+const Timestamp = styled.p`
+  color: #BECCE5;
```
kemuru, line 101: instead of this:

`color: #BECCE5;`

do this: 

`color: ${({ theme }) => theme.secondaryText};`

this is because we have all the color scheme centralized for both dark mode and light mode, you can check on the `styles/themes.ts` folder

kemuru:
need to add this change into the other core subgraphs too (core-neo, and core-university)

image

kemuru (approved):
lgtm, need to redeploy official core subgraphs

kemuru (approved):
lgtm

PR #1625, opened 2024-06-25
fix(web): prevent app crash while searching cases

nikhilverma360:
prevent app from crashing when we input special character ("/", "#")

reference :- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI#examples. (example section)

---

## PR-Codex overview
The focus of this PR is to update the encoding and decoding methods for URI filters in `uri.ts`.

### Detailed summary
- Updated `encodeURI` to `encodeURIComponent` for filter encoding
- Updated `decodeURI` to `decodeURIComponent` for filter decoding

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved URL encoding accuracy for filter objects by switching to `encodeURIComponent`, ensuring better handling of special characters.


alcercu (approved):
lgtm

PR #1781, opened 2024-12-10
refactor: remove indicators from voting tab, move to overview tab, su…

kemuru:
…bstitute current verdict compon


alcercu (approved):
lgtm

PR #1796, opened 2024-12-16
Chore/court stats no accordion

jaybuidl:
The accordion is not necessary since the statistics are occupying their own column anyways, even on mobile imo.

image


jaybuidl:
Done in https://github.com/kleros/kleros-v2/pull/1826

PR #1782, opened 2024-12-11
RNG Fallback

jaybuidl:
⚠️ This PR has been recreated as #2054 which replaces it.

---

This PR is a redo of [#966](https://github.com/kleros/kleros-v2/pull/966) as it has deviated too much from dev. It is based atop the `feat/chainlink-vrf-as-rng2` branch (PR #1778).

This PR introduces a composite RNG called `RNGWithFallback` responsible for the fallback logic. 

This PR breaks the storage layout of the SortitionModule so it cannot be upgraded and will require a full redeploy.

The RNGs are permissioned to prevent 3rd parties from using up the funds on the RNG oracle accounts with Chainlink or Randomizer.

The RNGs do not support concurrent requests, it is not necessary since only 1 consumer is allowed (the sortition module), with the notable exception of some governance function triggering a re-request, which invalidates the previous one.

This PR introduces several ABI changes:
- RNG interface renamed IRNG
- Removal of the _block parameter from the IRNG interface which is not necessary for VRFs, only for the BlockHashRNG which now tracks the block numbers internally

---

## PR-Codex overview
This PR focuses on refactoring the random number generation system by introducing the `IRNG` interface, modifying contracts to utilize this interface, and updating deployment scripts accordingly. It enhances flexibility and standardizes random number generation across various implementations.

### Detailed summary
- Deleted `RNG.solcspell.json`.
- Introduced `IRNG` interface in `contracts/src/rng/IRNG.sol`.
- Updated `SortitionModule` and `SortitionModuleNeo` to use `IRNG` instead of `RNG`.
- Changed `RNG_LOOKAHEAD` to `RNG_LOOKAHEAD_TIME` in deployment scripts.
- Modified `IncrementalNG`, `RandomizerRNG`, `ChainlinkRNG`, and `BlockHashRNG` to implement `IRNG`.
- Adjusted function signatures in RNG contracts to remove block parameters.
- Updated test cases to reflect changes in randomness request and receipt methods.
- Added `RNGWithFallback` contract to provide a fallback mechanism for RNG requests.

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

## Summary by CodeRabbit

* **New Features**
  * Introduced a modular random number generator (RNG) interface for improved flexibility.
  * Added support for Chainlink VRF and fallback RNG mechanisms.
  * Enhanced role management with distinct governor and consumer roles for RNG contracts.

* **Improvements**
  * Simplified and unified RNG contract interfaces for easier integration.
[...]

PR #966, opened 2023-06-21
RNG: Chainlink VRF + Fallback mechanism

zmalatrax:
This PR implements the Chainlink VRF Mechanism using the Subscription Method.

# Rationale Subscription Method vs Direct Funding
## Cost Efficiency
The Subscription Method reduces the gas cost of using VRF by pre-funding the subscription account with LINK when the direct funding method needs to calculate the cost of the request and fund it before actually making the request.

It is particularly efficient when many requests are made, which is the case as a random number is generated whenever jurors are drawn.

## Simplified Management
Using a subscription centralizes the management:
- A UI is provided to monitor the different requests made and their state.
- LINK tokens funding is directed toward a single contract rather than each consumer (in case of having multiple consumers, i.e. multiple functionality where random number generation is needed).
- The same applies to withdrawing. Withdrawing LINK tokens only has to be done from the subscription, not the consumers.

## Scalability
One subscription can support up to 100 consuming contracts. It allows for potential future expansion, which may require random number generation.

# Contracts
## [VRFConsumerV2.sol](https://github.com/kleros/kleros-v2/blob/feat/chainlink-vrf-as-rng/contracts/src/rng/VRFConsumerV2.sol)
VRF Consumer tailored to the SortitionModule.

It inherits from [VRFConsumerBaseV2.sol](https://github.com/kleros/kleros-v2/blob/feat/chainlink-vrf-as-rng/contracts/src/rng/VRFConsumerBaseV2.sol), provided by [Chainlink](https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/vrf/VRFConsumerBaseV2.sol)

Compared to the [RandomizerRNG](https://github.com/kleros/kleros-v2/blob/feat/chainlink-vrf-as-rng/contracts/src/rng/RandomizerRNG.sol), only the [SortitionModule](https://github.com/kleros/kleros-v2/blob/feat/chainlink-vrf-as-rng/contracts/src/arbitration/SortitionModule.sol) can make a request, and the callback function calls `sortitionModule.passPhase()` once the random number is ready.

Implementation doc: [Chainlink - Get a random number](https://docs.chain.link/vrf/v2/subscription/examples/get-a-random-number)

## [VRFSubscriptionManagerV2.sol](https://github.com/kleros/kleros-v2/blob/feat/chainlink-vrf-as-rng/contracts/src/rng/VRFSubscriptionManagerV2.sol)
A contract allowing programmatic management of the VRF subscription.

[...]

PR #2054, opened 2025-07-24
RNG Fallback

jaybuidl:
This PR is a redo of #966 and #1782 as they had deviated too much from dev. It was originally based atop the `feat/chainlink-vrf-as-rng2` branch (PR #1778).

This PR introduces a `RNGWithFallback` responsible for the fallback logic. 

We ensure that repeated calls to `SM.passPhase()` does not result in repeated calls to the default RNG or the fallback RNG as it would keep overwriting the `requestID` for VRF-based RNGs or the `requestBlock` for the BlockhashRNG.

The RNGs are permissioned to prevent 3rd parties from using up the funds on the RNG oracle accounts with Chainlink or Randomizer.

The RNGs do not support concurrent requests, it is not necessary since only 1 consumer is allowed (the sortition module), with the notable exception of some governance function triggering a re-request, which invalidates the previous one.

We rely on `block.timestamp` rather than `block.number` because it is a more reliable measure of the elapsed time on Arbitrum, as block production is sporadic depending on network conditions.

### ABI changes
- `RNG` interface renamed `IRNG`
- Removal of the `_block` parameter from the `IRNG` interface which is not necessary for VRFs, only for the BlockHashRNG which now tracks it internally

---

## PR-Codex overview
This PR focuses on refactoring the random number generator (RNG) implementation in the Kleros project, introducing a new interface `IRNG`, and enhancing the RNG architecture with fallback mechanisms.

### Detailed summary
- Deleted `RNG.sol` and related deployment scripts.
- Introduced `IRNG` interface for RNG functionalities.
- Created `RNGWithFallback` contract for fallback mechanisms.
- Updated `IncrementalNG`, `ChainlinkRNG`, and `RandomizerRNG` to implement `IRNG`.
- Removed unnecessary parameters from RNG functions.
- Adjusted consumer and governor access controls in RNG contracts.
- Updated tests to reflect changes in RNG calls and expectations.

> The following files were skipped due to too many changes: `contracts/test/foundry/KlerosCore.t.sol`

## Summary by CodeRabbit

* **New Features**
  * Added a fallback RNG with readiness checks and observable fallback events for more reliable randomness.

* **Refactor**
  * Unified RNG interface and simplified RNG–sortition interaction; removed block/lookahead timing parameters.

* **Chores**
  * Updated deployment wiring to use the fallback RNG and removed legacy RNG deployment scripts.

* **Tests**
  * Migrated tests to parameterless request/ready/read flow, timestamp-based pacing, and added fallback tests.

* **Documentation**
  * Updated changelog with breaking changes and RNG behavioral notes.


unknownunknown1:
Updated the RNG contracts according to [my own feedback](https://docs.google.com/document/d/1GwV1xRtnIVxGR4RkWQ-pTMI7r4rPavjPHkSDKEJlZ0A/edit?tab=t.0)
I ended up making only partial changes because implementing them fully ended up being problematic:

- BlockhashRNG contract was remained untouched, only a couple of clarifying comments was added
- I attempted to implement an upper bound for the timestamp. The idea was to reset `requestTimestamp` when that bound was exceeded. The problem with this is that when you reset the timestamp RNG returns `0` and passes it to [SortitionModule](https://github.com/kleros/kleros-v2/blob/e3ed3c95030d2f23e6307942020f11521fd60767/contracts/src/arbitration/SortitionModuleBase.sol#L184), which makes it revert, thus unless we want to change the logic in SM (which I don't think is warranted) it's not possible to change storage during unsuccessful random number reception. I considered implementing the upper bound using the logic similar to the one used in  [BeaconRNG](https://github.com/kleros/kleros-interaction/blob/faf6faa53d39a106837f8ca22f12d2ae5227bd41/contracts/standard/rng/BeaconRNG.sol#L51) but it only seemed to add unnecessary complications.
- For the aforementioned issue the fallback contract was greatly simplified. That issue was present in the previous version too, but wasn't spotted during the first review. So since it's not possible to change storage when `receiveRandomness` fails I implemented an immediate fallback based on BlockhashRNG logic
- Recent changes were causing stack-too-deep error during Foundry compilation, which this PR fixes

jaybuidl:
@coderabbitai resolve

PR #2277, opened 2026-04-22
feat: pre-commit type checking

kyrers:
Resolves #2273.

---

## PR-Codex overview
This PR primarily focuses on enhancing TypeScript type definitions, improving error handling, and updating linting configurations across multiple files. It also makes several adjustments to ensure better type safety and code readability.

### Detailed summary
- Added TypeScript type declarations for `@reality.eth/reality-eth-lib`.
- Changed `stake` type from `string` to `bigint` in multiple interfaces.
- Improved error handling by removing unnecessary catch parameters.
- Updated linting configurations to include additional rules.
- Refactored code for better type safety and readability in various components.
- Adjusted function signatures to use more specific types.
- Ensured consistent use of `Address` type across components.

[...]

PR #2279, opened 2026-05-06
feat: pre-commit type checking and lint - web

kyrers:
Part of #2277, but isolates the `web` workspace changes for easier review.

---

## PR-Codex overview
This PR focuses on various code improvements, including type adjustments, error handling enhancements, and updates to the styling and structure of components. It aims to improve type safety, maintainability, and overall code quality.

### Detailed summary
- Updated `stake` type from `string` to `bigint` in several components.
- Enhanced error handling by removing unnecessary error variable in `catch` statements.
- Added TypeScript type declarations for various modules.
- Adjusted transport type in `client` configuration.
- Improved consistency in handling `Address` types across components.
- Refactored `useCoinPrice` to accept an array of coin IDs.
- Simplified JSX structures and improved readability in multiple components.
- Updated the way `numberOfJurors` is handled to avoid type issues.
- Enhanced linting rules and configurations for better code quality.
- Made various styling adjustments for consistency and clarity in UI components.

> The following files were skipped due to too many changes: `web/src/pages/Jurors/index.tsx`, `web/src/pages/Profile/Votes/VoteCard/index.tsx`, `web/src/pages/Profile/JurorCard/BottomContent/index.tsx`, `web/src/hooks/useHomePageContext.tsx`, `web/src/components/JurorLink.tsx`, `web/src/utils/getDrawnJurorsWithCount.ts`, `web/src/utils/crypto/hashJustification.ts`, `web/src/tests/e2e/utils/contracts.ts`, `web/src/components/Verdict/FinalDecision.tsx`, `web/src/hooks/queries/useAppealCost.ts`, `web/src/components/TxnHash.tsx`, `web/src/layout/Footer/index.tsx`, `web/src/layout/Header/navbar/Menu/Settings/Notifications/FormContactDetails/index.tsx`, `web/src/pages/Profile/Cases/index.tsx`, `web/src/pages/Home/TopJurors/JurorCard/MobileCard.tsx`, `web/src/pages/Cases/CaseDetails/Timeline.tsx`, `web/src/pages/Resolver/Parameters/Jurors.tsx`, `web/src/pages/Profile/Stakes/CurrentStakes/index.tsx`, `web/src/pages/Cases/CaseDetails/Voting/VotesDetails/index.tsx`, `web/src/components/DisputeView/DisputeInfo/DisputeInfoList.tsx`, `web/src/pages/Home/TopJurors/JurorCard/DesktopCard.tsx`, `web/src/utils/getDisputeRequestParamsFromTxn.ts`, `web/src/hooks/queries/useDisputeKitClassicMultipliers.ts`, `web/src/layout/Header/navbar/Menu/Settings/Notifications/index.tsx`, `web/src/components/DisputeView/DisputeListView.tsx`, `web/src/components/DisputeView/DisputeCardView.tsx`, `web/src/components/DisputePreview/DisputeContext.tsx`, `web/src/pages/Courts/CourtDetails/StakingHistoryByCourt/DisplayStakes/index.tsx`, `web/src/components/GradientTokenIcons.tsx`, `web/src/pages/Cases/CaseDetails/Voting/VotingHistory.tsx`, `web/src/pages/Jurors/DisplayJurors.tsx`, `web/src/pages/Home/CourtOverview/Chart.tsx`, `web/src/components/ClaimPnkButton.tsx`, `web/src/components/DisputeView/CardLabels/index.tsx`, `web/src/hooks/queries/usePolicyRegistryEvent.ts`, `web/src/pages/Cases/CaseDetails/Voting/Classic/Vote.tsx`, `web/src/pages/Courts/CourtDetails/TopSearch.tsx`, `web/src/pages/Profile/JurorCard/index.tsx`, `web/src/hooks/queries/usePopulatedDisputeData.ts`, `web/src/pages/Profile/Votes/index.tsx`, `web/src/pages/Courts/CourtDetails/StakePanel/StakeWithdrawButton.tsx`, `web/src/pages/Home/CourtOverview/Stats.tsx`, `web/src/pages/Profile/JurorCard/StakingRewards.tsx`, `eslint-config/README.md`

## Summary by CodeRabbit

* **New Features**
  * Two new Home Court Overview charts: Cases by Courts and Staked PNK by Courts.
  * Product menu icons show a circular skeleton placeholder while image assets load.
  * Swap success popup subtitle standardized for non-claim swaps.

* **Chores**
  * Added pre-commit type checks for the web workspace.
  * Enabled linting and Prettier formatting on staged JS/TS files.
  * Tightened TypeScript settings to enforce stricter checks and include source files.


kyrers:
@coderabbitai review

web/src/actions/commit/builders/classic.builder.ts
```diff
@@ -17,9 +17,11 @@ export const classicCommitBuilder = defineCommitBuilder({
     deps.storeCommitData(key, { choice, salt });
 
     const commit = hashVote(choice, salt);
+    const chainKey = chain.id as keyof typeof disputeKitClassicAddress;
+
     return {
       account,
-      address: disputeKitClassicAddress[chain.id],
+      address: disputeKitClassicAddress[chainKey],
       abi: disputeKitClassicAbi,
```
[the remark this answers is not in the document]
kyrers: Behavior is an edge case and preexisting; this change simply fixes the type issue. Out of scope for this PR.

web/src/utils/getDrawnJurorsWithCount.ts
```diff
@@ -1,9 +1,11 @@
+import { Hash } from "viem";
```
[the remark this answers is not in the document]
kyrers: As acknowledged, the proper fix lies elsewhere. Also, this is an edge case. These strings come from the subgraph and are indexed from the blockchain, so validity is expected.

web/src/pages/Resolver/Parameters/Jurors.tsx
```diff
@@ -62,11 +62,12 @@ const Jurors: React.FC = () => {
...
     },
     args: [prepareArbitratorExtradata(disputeData.courtId ?? "", disputeData?.numberOfJurors ?? 0)],
-    chainId: DEFAULT_CHAIN.id,
+    chainId: DEFAULT_CHAIN.id as keyof typeof klerosCoreAddress,
   });
 
   const arbitrationFee = formatETH(data ?? BigInt(0), 18);
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   useEffect(() => setDisputeData({ ...disputeData, arbitrationCost: data?.toString() }), [data]);
```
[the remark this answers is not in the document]
kyrers: The effect's only meaningful dependency is data. When it changes, we overwrite arbitrationCost. Spreading `disputeData` is harmless, and switching to a functional updater would require changes in `useLocalStorage` and `useNewDisputeContext`, thus out of scope for this PR.

web/src/pages/Cases/CaseDetails/Voting/Classic/Commit.tsx
```diff
@@ -4,6 +4,8 @@ import styled from "styled-components";
 import { useParams } from "react-router-dom";
 import type { Address } from "viem";
 
+import { ClassicCommitParams } from "~src/actions/commit/params";
```
tractorss: `~`

web/src/pages/Cases/CaseDetails/Voting/Classic/Commit.tsx
```diff
@@ -48,7 +50,7 @@ const Commit: React.FC<ICommit> = ({ arbitrable, voteIDs, setIsOpen, disputeKitN
         choice,
         voteIds: parsedVoteIDs,
         roundIndex: Number(currentRoundIndex),
-      });
+      } as ClassicCommitParams);
```
tractorss: This is narrowing the type to `ClassicCommitParams` while the actual type is `PartialBy<CommitParams, "salt">` 

I can't find my note on this specific issue, but to put a rough explanation:
Typescript discriminated unions work well upto 25 members. And as soon as it passes that, it becomes too complex for the compiler to check the types against this Discriminated union.

In our case, you may notice that CommitParams is just a union of 5, so why the error. Issue comes when in `useCastCommit` with do this:

`type CommitVoteParams = PartialBy<CommitParams, "salt">;`

This is done to overcome another issue with how typescript's Omit collapses the Discriminated Unions and  looses the strucuture, so we needed that to maintain the structure while using `Omit`.

Now this new type `CommitVoteParams` is a discriminated union of >25, (5^2). And hence ts compiler throws error when we assign a single object to it, because it can't compare them.

That is what I was able to infer from my searches, but maybe there's another underlying issue here.
reference: https://stackoverflow.com/questions/77144586/typescript-intersecting-unions-stop-working-after-5-guard-properties

Solution: for now assert the passed object in `castCommit` as `PartialBy<CommitParams, "salt">``

web/src/pages/Cases/CaseDetails/Voting/Classic/Vote.tsx
```diff
@@ -37,16 +41,20 @@ const Vote: React.FC<IVote> = ({ arbitrable, voteIDs, setIsOpen, disputeKitName
...
       await vote({
         disputeId: parsedDisputeID,
         voteIds: parsedVoteIDs,
         choice: voteOption,
-        salt: BigInt(disputeData?.dispute?.currentRoundIndex),
+        salt: BigInt(currentRoundIndex),
         justification,
         type: disputeKitName ?? DisputeKits.Classic,
-      });
+      } as ClassicVoteParams);
```
tractorss: Same as above

web/src/pages/Cases/CaseDetails/Voting/VotesDetails/index.tsx
```diff
@@ -194,7 +199,9 @@ const VotesAccordion: React.FC<IVotesAccordion> = ({ drawnJurors, period, answer
     <>
       {drawnJurors.length === 0 ? <StyledInfoCard msg={t("alerts.jurors_not_drawn_yet")} /> : null}
       <Container>
-        {accordionItems.length > 0 ? <StyledAccordion items={accordionItems} /> : null}
+        {accordionItems.length > 0 ? (
+          <StyledAccordion items={accordionItems as unknown as CustomAccordionItem[]} />
```
tractorss, line 203: As we discussed, this usage is dangerous as there's no real type check.

I remember for TimelineItems we do something like
`type TimelineItems = [TimelineItem, ...TimelineItem[]]`
Maybe we can do something similar here?

web/src/pages/Cases/CaseDetails/Tabs.tsx
```diff
@@ -84,9 +84,10 @@ const Tabs: React.FC = () => {
 
   const tabs = useMemo(() => {
     const updatedTabs = [...TABS];
+    const periodIndex = parseInt(String(currentPeriodIndex));
```
tractorss: Sonar cloud seems correct

web/src/pages/Home/TopJurors/Header/Rewards.tsx
```diff
@@ -10,6 +10,8 @@ import { landscapeStyle } from "styles/landscapeStyle";
 import WithHelpTooltip from "components/WithHelpTooltip";
 
 const Container = styled.div<{ shortLabel: string; longLabel: string }>`
+  --label-long: "${({ longLabel }) => longLabel}";
```
tractorss, line 13: I like this approach, any specific reason to add it 👀

web/src/pages/Profile/Cases/index.tsx
```diff
@@ -47,8 +48,9 @@ const Cases: React.FC<ICases> = ({ searchParamAddress }) => {
   );
 
   const { data: userData } = useUserQuery(searchParamAddress, decodedFilter);
+  const userDetails = readFragment(userFragment, userData?.user);
```
tractorss, line 51: If i understand correctly, given a complete user data it will only read and return the data pertaining to `userFragment` query, with proper type?

web/src/pages/Profile/Votes/VoteCard/CaseStatus.tsx
```diff
@@ -3,12 +3,14 @@ import styled, { css, useTheme } from "styled-components";
 
 import { useTranslation } from "react-i18next";
 
+import { Period } from "~src/graphql/graphql";
```
tractorss: `~`

tractorss: on a second thought, can you check if we can make the linter check for this kind of imports, or change the relative import structure

web/src/pages/Resolver/Landing/index.tsx
```diff
@@ -120,7 +120,9 @@ const Landing: React.FC = () => {
       policyURI: populatedDispute.policyURI,
       question: populatedDispute.question,
       courtId: roundData.round?.court.id,
-      numberOfJurors: roundData.round?.nbVotes,
+      //changing the subgraph schema for Round.nbVotes from BigInt to Int
+      //will automattically resolve to number and we can remove this parse
```
tractorss: hmm, do we need too 🤔 , we usually treat the direct values from on-chain as `bigint`, and round.nbVotes is directly saved from on chain data in subgraph

web/src/pages/Resolver/Preview/index.tsx
```diff
@@ -79,7 +80,9 @@ const Preview: React.FC = () => {
       <StyledCard>
         <Overlay />
         <PreviewContainer>
-          <DisputeContext disputeDetails={disputeTemplate} />
+          {/*IDisputeTemplate is a subset of DisputeDetails.
```
tractorss, line 83: `DisputeDetails` from the sdk is the populated dispute type and doesn't include dispute template.

IDisputeTemplate is a separate type?

web/src/utils/getDrawnJurorsWithCount.ts
```diff
@@ -19,7 +21,9 @@ export const getDrawnJurorsWithCount = (drawnJurors: IVotingHistoryRounds) =>
           juror: { id: jurorId },
           voteCount: 1,
           vote: current.vote,
-          transactionHash: current.vote?.justification?.transactionHash,
+          //once the subgraph schema for ClassicJustification.transactionHash changes from String! to Bytes!
+          //it will automattically resolve to 0xstring and we can remove this cast
```
tractorss, line 25: let's track this in an issue, Ideally one issue with sub-issues.

web/src/hooks/useHomePageContext.tsx
```diff
@@ -6,23 +6,21 @@ export type { HomePageQuery };
...
 });
 
 export const HomePageProvider: React.FC<{
   children: React.ReactNode;
   timeframe: number;
 }> = ({ children, timeframe }) => {
-  const { data, error, isValidating } = useHomePageQuery(timeframe);
-  return <Context.Provider value={{ data, error, isValidating }}>{children}</Context.Provider>;
+  const { data, error } = useHomePageQuery(timeframe);
+  return <Context.Provider value={{ data, error }}>{children}</Context.Provider>;
```
tractorss: Since sonar cloud is complaining, let's add it here. Although bit off PR topic, but is a small change

tractorss (commented):
looks great, left a couple comments 🙌

web/src/pages/Cases/CaseDetails/Voting/Classic/Commit.tsx
kyrers: Nice catch. I always miss some of these. Addressed in 5d37038505552ceb92f6c6e653e1e44fb4b17caf.

web/src/pages/Profile/Votes/VoteCard/CaseStatus.tsx
kyrers: Fixed and rule added to ESLint config on 5d37038505552ceb92f6c6e653e1e44fb4b17caf.

web/src/pages/Cases/CaseDetails/Voting/Classic/Commit.tsx
kyrers: Thank you for that explanation, good to know. I used the cast to `ClassicCommitParams` because this was in the Classic voting component. That said, you're indeed right that it expects `PartialBy<CommitParams, "salt">`, so it's better to use that. Addressed in 016f52a26437fb25fa9b8677150b96507cc5efea.

web/src/pages/Cases/CaseDetails/Voting/Classic/Vote.tsx
kyrers: I addressed this in 016f52a26437fb25fa9b8677150b96507cc5efea. However, here, it seems the function does not expect partial types. It expects `VoteParams`, so I used that.

web/src/pages/Cases/CaseDetails/Voting/VotesDetails/index.tsx
kyrers: So, two things here:

1. We cannot do something similar to `TimelineItems`, because of [this bug](https://github.com/kleros/kleros-v2/pull/2279/changes#diff-725407cf039b5aecd27c59f73e6a1135a5c8cfb0faa386d740bfe16cf17b0d63R159-R163) with the version of the library used. Once we update it, we can do this, provided the bug does not exist in newer versions of the library.

2. While not ideal, this is not totally unsafe as we build the accordion items ourselves. So, I used `unknown` to get past the type error, knowing that the proper fix is only possible once the component library is updated.

web/src/pages/Cases/CaseDetails/Tabs.tsx
kyrers: Totally, my bad. Addressed in 03850cb4eff35c9779894d1668160a0741e3e4b7.
Note that this happens in many places and is out of scope for this PR (IMO), but we should keep a note of doing this replacement anywhere we spot it going forward.
Alternatively, I can do it in this PR if you think it's not scope creep.

web/src/pages/Home/TopJurors/Header/Rewards.tsx
kyrers: Yeah, so, the original code tried to use `longLabel` inside the `landscapeStyle(() => css...` block, but this inner function doesn't seem to have access to the `styled-component` props. So `longLabel` just got stringified into garbage, and that's what was set on `content`. Luckily for us, this is not really visible, so we never noticed it. 
However, with the type checks, this became flagged as a type problem. The easier solution was just to interpolate it at the top and then just use it as a css var, which works everywhere.

web/src/pages/Profile/Cases/index.tsx
kyrers: Yup, exactly. I am not sure it even needs the complete user data, as long as the data sent has enough overlap with the `userFragment` in this case. 
Note, however, that this is originally named `useFragment`, I aliased it to `readFragment` on import for two reasons:
1. To avoid false rules-of-hooks violations because of the `use` prefix.
2. We're not hooking into anything, we are reading part of a fragment.

web/src/pages/Resolver/Landing/index.tsx
kyrers: No, we don't need to. I just double checked, and I think I made a mistake. I added this comment because I was under the impression this was not going to be a `uint256` anymore, in which case doing this would be accurate and allow us to not need the parse. I must have confused it with something else I saw on the contracts.

Anyway, if indeed it will remain `uint256`, my suggestion would just mismatch the types for the sake of one parse, so might as well keep the parse. Removed the comment in bad5273638726e04513000fbebc5b87140676dc1.

web/src/pages/Resolver/Preview/index.tsx
kyrers: Exactly, defined locally in [NewDisputeContext](https://github.com/kleros/kleros-v2/blob/feat/pre-commit-type-check-web/web/src/context/NewDisputeContext.tsx#L37). Hence my comment that the proper fix is consolidating the types, but it seems like a bigger change and out of scope for this PR.

web/src/hooks/useHomePageContext.tsx
kyrers: Addressed in 0b71e537ff931ad7347c659663c148877d100542.

web/src/utils/getDrawnJurorsWithCount.ts
kyrers: @coderabbitai, create an issue to track this. Doesn't need to have or be a sub-issue for now. Explain the context. No need to assign it to anyone either.

web/src/pages/Cases/CaseDetails/Voting/VotesDetails/index.tsx
kyrers: @coderabbitai, create an issue for this too. Once the library is updated, we should be able to simplify this. Note in the issue that even after updating the library, if we're unable to simplify this, it should be fixed in the library ASAP.

kyrers:
@coderabbitai, also create an issue for [this](https://github.com/kleros/kleros-v2/pull/2279/changes#diff-11128dc71632f186aa0629e634663d68a1c15a5893781c7d77d2b8d7af168e42). It was added as a shim to prevent type errors. However, soon we will be migrating the UI to Tailwind, at which point we should also replace this library. Create an issue for replacing this library, adding both why and when it should happen.

web/src/pages/Cases/CaseDetails/Voting/VotesDetails/index.tsx
tractorss: sounds good

web/src/pages/Resolver/Preview/index.tsx
tractorss: okay I see, yea the name is conflicting with the general templates we have, here it is more of a template to build upon the dispute in resolver.

tractorss (approved):

web/src/pages/Home/TopJurors/Header/Rewards.tsx
tractorss: nice catch !

web/src/pages/Cases/CaseDetails/Tabs.tsx
tractorss: I agree, it's also not big issue. We can do as you say, with updating as we see the uses 👍

PR #2106, opened 2025-08-23
feat: add keyboard navigation for court search results

ManiBAJPAI22:
##  **Background**

Adds accessible keyboard navigation support for the court search results in the Courts page search bar.

##  **New Features**

- **ArrowUp/ArrowDown**: Navigate through search results with visual feedback
- **Enter Key**: Select the highlighted result and navigate to that court
- **Visual Highlighting**: Selected items get blue left border and background styling
- **Smart Selection**: Current court only shows as selected when no keyboard navigation is active

##  **Technical Changes**

- Added `selectedIndex` state to track keyboard navigation
- Implemented `handleKeyDown` for ArrowUp/ArrowDown/Enter key handling
- Updated `StyledCard` component to support `keyboardSelected` prop
- Added proper event prevention and boundary checking
- Maintained all existing mouse navigation functionality

##  **Testing**

- ✅ ArrowUp/ArrowDown navigation works correctly
- ✅ Enter key selects and navigates to highlighted results
- ✅ Visual highlighting appears for keyboard-selected items
- ✅ Mouse clicks continue to work as before
- ✅ Selection state resets properly on search changes

##  **Files Changed**

- `web/src/pages/Courts/CourtDetails/TopSearch.tsx` - Added keyboard navigation logic and styling

## ♿ **Accessibility Impact**

This change significantly improves keyboard accessibility for users who rely on keyboard navigation, making the court search functionality more inclusive and user-friendly.

## 🎯 **What does this PR do?**

Adds accessible keyboard navigation support for the court search results in the Courts page search bar.

## ✨ **New Features**

- **ArrowUp/ArrowDown**: Navigate through search results with visual feedback
- **Enter Key**: Select the highlighted result and navigate to that court
- **Visual Highlighting**: Selected items get blue left border and background styling
- **Smart Selection**: Current court only shows as selected when no keyboard navigation is active

## 🔧 **Technical Changes**

- Added `selectedIndex` state to track keyboard navigation
- Implemented `handleKeyDown` for ArrowUp/ArrowDown/Enter key handling
- Updated `StyledCard` component to support `keyboardSelected` prop
- Added proper event prevention and boundary checking
- Maintained all existing mouse navigation functionality

## 🧪 **Testing**

- ✅ ArrowUp/ArrowDown navigation works correctly
- ✅ Enter key selects and navigates to highlighted results
- ✅ Visual highlighting appears for keyboard-selected items
- ✅ Mouse clicks continue to work as before
- ✅ Selection state resets properly on search changes

[...]

Issue #1823, opened 2025-01-08
Accessibility backlog: Add keyboard navigation


alcercu:
_:hammer_and_wrench: Refactor suggestion_

**Add keyboard navigation for search results.**

The search results can only be navigated using a mouse. Consider adding keyboard navigation for better accessibility.

```diff
+  const [selectedIndex, setSelectedIndex] = useState(-1);
+
+  const handleKeyDown = (e: React.KeyboardEvent) => {
+    if (!search || filteredCourts.length === 0) return;
+    
+    switch(e.key) {
+      case 'ArrowDown':
+        e.preventDefault();
+        setSelectedIndex(prev => Math.min(prev + 1, filteredCourts.length - 1));
+        break;
+      case 'ArrowUp':
+        e.preventDefault();
+        setSelectedIndex(prev => Math.max(prev - 1, -1));
+        break;
+      case 'Enter':
+        if (selectedIndex >= 0) {
+          navigate(`/courts/${filteredCourts[selectedIndex].id}`);
+          setSearch("");
+        }
+        break;
+    }
+  };

   return (
     <SearchBarContainer>
       <StyledSearchbar
         dir="auto"
         type="text"
         placeholder="Search"
         value={search}
         onChange={(e) => setSearch(e.target.value)}
+        onKeyDown={handleKeyDown}
       />
       {search && filteredCourts.length > 0 && (
         <SearchResultsContainer>
           {filteredCourts.map((court, index) => (
             <StyledCard
               key={court.id}
-              selected={court.id === currentCourtId}
+              selected={court.id === currentCourtId || index === selectedIndex}
               onClick={() => {
                 navigate(`/courts/${court.id}`);
                 setSearch("");
               }}
             >
```

> Committable suggestion skipped: line range outside the PR's diff.

_Originally posted by @coderabbitai[bot] in https://github.com/kleros/kleros-v2/pull/1821#discussion_r1906391340_

Issue #98, opened 2022-06-20
Allow Dispute Kits to unstake ineligible jurors


jaybuidl:
## Problem
For the 1 Human 1 Vote Dispute Kit, it is possible that some large staked jurors might not be registered on Proof of Humanity. In such a case, they would end up being drawn often from the sortition tree but later filtered out. It is highly inefficient and costly.

## Proposed solution
Some team members have suggested that KlerosCore should allow a DisputeKit to unstake such ineligible jurors as follow:

https://github.com/kleros/kleros-v2/blob/51ec1d2894560781371fa8ca97e2b5f08311bed3/contracts/src/arbitration/dispute-kits/DisputeKitSybilResistant.sol#L233-L241

https://github.com/kleros/kleros-v2/blob/51ec1d2894560781371fa8ca97e2b5f08311bed3/contracts/src/arbitration/KlerosCore.sol#L481-L487

#### A) If there are 2 Dispute Kits with different eligibility criteria using the same subcourt
Then one Dispute Kit might decide to unstake a juror while this juror is still eligible in the other Dispute Kit. It would be unfair to the juror as he would miss some opportunities. There is no easy way to prevent that.

#### B) Should unstaking ineligible jurors take place in the usual Staking phase?
- Yes because that's what phases are made for. But then the juror could re-stake immediately after.
- No and by-pass the DelayedStakes to prevent the juror from re-staking immediately after, but this might still happen if it takes more than maxDrawingTime to draw and KlerosCore cycles through the phases. In addition it breaks the design of the phases and creates a new path for unstaking.

jaybuidl:
This issue should be viewed as an optimisation which benefits the bot operators. It does not benefit the end users with either better functionality or performance.

For now we are waiting for more clarity on the possibility of making the sortition sum tree more modular + the possibility of simplifying the phases in Kleros Core.

jaybuidl:
#137 has been implemented so we can revisit this.

Unstake ineligible jurors from court X:
- only during Staking phase
- and only if they are ineligible from all the DisputeKits allowed by the court in question.

![image](https://user-images.githubusercontent.com/22213980/234571162-86b2d0c8-824d-4430-8eec-ac5a396c5cd3.png)

jaybuidl:
Currently we cannot iterate through the dispute kits to implement this logic.

Worst case just unstake without checking all the DKs.

jaybuidl:
Easier to implement now after #1357

PR #2281, opened 2026-05-15
feat: migration to latest ui components library with file viewer

kemuru:
## Summary by CodeRabbit

* **New Features**
  * Updated the interface to the latest UI components, including refreshed dropdowns, fields, tabs, radio controls, buttons, and timelines.
  * Added accessible labels for search fields and selection controls in English, Spanish, and French.
  * Added improved ETH amount input formatting and validation.
  * Added dark-mode styling synchronization and refreshed typography.

* **Improvements**
  * Improved responsive layouts, spacing, link presentation, and tooltip readability.
  * Notifications navigation now uses query parameters instead of URL hashes.

* **Changes**
  * Removed built-in document viewer support for Markdown, SVG, and other local file rendering.

---

## PR-Codex overview
This PR focuses on updating various components and styles across the application, primarily enhancing button interactions and refining layout styles. It also involves migrating to new components from the `@kleros/ui-components-library`.

### Detailed summary
- Deleted several unused components.
- Updated button interactions from `onClick` to `onPress`.
- Introduced `StyledInternalLink` with new styles.
- Enhanced layout with grid and flex properties.
- Revised form fields to utilize `TextField` instead of `Field`.
- Adjusted margin and padding for better spacing.
- Improved accessibility features with `aria-label` attributes.
- Updated theme colors and styles for consistency.
- Refactored components to use new styled components.

> The following files were skipped due to too many changes: `web/src/pages/Resolver/Landing/index.tsx`, `web/src/pages/Courts/CourtDetails/Description.tsx`, `web/src/tests/e2e/voting.spec.ts`, `web/src/pages/AttachmentDisplay/index.tsx`, `web/src/layout/Header/navbar/Menu/Settings/Notifications/FormContactDetails/EmailVerificationInfo.tsx`, `web/src/pages/Resolver/Parameters/VotingOptions/OptionsFields.tsx`, `web/src/pages/Resolver/Parameters/NotablePersons/PersonFields.tsx`, `web/src/hooks/useBigNumberFieldReset.test.tsx`, `web/src/components/DisputeFeatures/Features/index.tsx`, `web/src/pages/Profile/index.tsx`, `web/src/pages/Home/CourtOverview/Chart.tsx`, `web/src/components/LabeledDropdown.tsx`, `web/src/styles/base-elements.css`, `web/src/pages/Courts/CourtDetails/TopSearch.tsx`, `web/src/pages/Cases/CaseDetails/Appeal/Classic/Options/StageOne.tsx`, `web/src/pages/Cases/CaseDetails/Tabs.tsx`, `web/src/pages/Resolver/Parameters/Court/index.tsx`, `web/src/components/CasesDisplay/Search.tsx`, `web/src/layout/Header/navbar/Menu/Settings/index.tsx`, `web/src/components/DisputeFeatures/Features/GatedErc20.tsx`, `web/src/layout/Header/DesktopHeader.tsx`, `web/src/pages/Profile/Votes/StatsAndFilters/Filters.tsx`, `web/src/pages/Resolver/Parameters/Jurors.tsx`, `web/src/tests/e2e/case.spec.ts`, `web/src/pages/Cases/CaseDetails/Appeal/Classic/Fund.tsx`, `web/src/components/CasesDisplay/Filters.tsx`, `web/src/pages/Courts/CourtDetails/StakePanel/StakeWithdrawPopup/stakeSteps.tsx`, `web/src/components/DisputeFeatures/Features/GatedErc1155.tsx`, `web/src/components/Verdict/DisputeTimeline.tsx`, `web/src/pages/Cases/CaseDetails/Appeal/OptionCard.tsx`, `web/src/pages/Courts/CourtDetails/StakePanel/InputDisplay.tsx`, `web/src/pages/Resolver/Landing/CreationCard.tsx`, `yarn.lock`


web/src/components/LightButton.tsx
```diff
@@ -38,14 +39,14 @@ const StyledButton = styled(Button)<{ isMobileNavbar?: boolean }>`
 interface ILightButton {
   text: string;
   Icon?: React.FC<React.SVGAttributes<SVGElement>>;
-  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+  onClick?: React.ComponentProps<typeof Button>["onClick"];
```
tractorss: in v3 library, Button now uses `onPress` which provides better accessibility

web/src/components/Radio.tsx
```diff
@@ -0,0 +1,52 @@
...
+interface RadioBaseProps {
+  small?: boolean;
+  checked?: boolean;
+}
+
+export interface RadioProps extends RadioBaseProps, Omit<React.InputHTMLAttributes<HTMLInputElement>, "type"> {
+  label: string;
+}
+
+const Radio: React.FC<RadioProps> = ({ label, small, className, ...props }) => (
```
tractorss: Use the Radio component exported by the library

web/src/layout/Header/navbar/Menu/Settings/index.tsx
```diff
@@ -74,23 +74,27 @@ const Settings: React.FC<ISettings> = ({ toggleIsSettingsOpen, initialTab }) =>
...
   return (
     <Container ref={containerRef}>
       <StyledSettingsText>{t("menu.settings")}</StyledSettingsText>
       <StyledTabs
-        currentValue={currentTab}
+        selectedKey={currentTab}
         items={TABS}
-        callback={(n: number) => {
-          setCurrentTab(n);
+        callback={(_key, value) => {
```
tractorss: iirc, v3 handles the tab state itself, so no need to maintain separately

web/src/pages/AttachmentDisplay/index.tsx
```diff
@@ -25,6 +25,8 @@ const Container = styled.div`
   margin: 0 auto;
 `;
 
+const FileViewerSurface = styled.div``;
```
tractorss: redundant?

web/src/pages/Cases/CaseDetails/Evidence/SubmitEvidenceModal.tsx
```diff
@@ -113,10 +113,10 @@ const SubmitEvidenceModal: React.FC<{
         variant="info"
       />
       <ButtonArea>
-        <Button variant="secondary" disabled={isSending} text={t("buttons.return")} onClick={close} />
+        <Button variant="secondary" isDisabled={isSending} text={t("buttons.return")} onClick={close} />
```
tractorss: replace `onClick` with `onPress`, same for all Button usages

web/src/pages/Cases/CaseDetails/Voting/VotingHistory.tsx
```diff
@@ -107,12 +107,14 @@ const VotingHistory: React.FC<{ arbitrable?: `0x${string}`; isQuestion: boolean
...
-              currentValue={currentTab}
+              selectedKey={currentTab}
               items={rounds.map((_, i) => ({
+                id: i,
                 text: t("voting.round_number", { number: i + 1 }),
                 value: i,
+                content: null,
               }))}
-              callback={(i: number) => setCurrentTab(i)}
+              callback={(_key, value) => setCurrentTab(value)}
```
tractorss, line 118: If not needed externally, we can let Tab handle the currentTab itself

web/src/pages/Cases/CaseDetails/Tabs.tsx
```diff
@@ -93,11 +101,11 @@ const Tabs: React.FC = () => {
 
   return (
     <StyledTabs
-      currentValue={currentTab}
+      selectedKey={currentTab}
```
tractorss, line 112: I don't think we need to pass this value again here, Tab should handle internally.
We can just consume the value outside

web/src/utils/uiComponentsTypes.ts
```diff
@@ -0,0 +1,25 @@
+import type { ComponentProps } from "react";
+
+import { CustomTimeline, DropdownCascader, DropdownSelect } from "@kleros/ui-components-library";
+
+/**
```
tractorss: explanation not necessary here

web/src/styles/themes.ts
```diff
@@ -1,7 +1,4 @@
-import { lightTheme as componentsLightTheme, darkTheme as componentsDarkTheme } from "@kleros/ui-components-library";
-
 export const lightTheme = {
-  ...componentsLightTheme,
```
tractorss, line 4: removal of this may have caused the change in colors that you noticed?

web/src/components/Radio.tsx
kemuru: v3 still exports Radio, but it now points to the RadioGroup component — there's no single-radio component behind that name anymore. So import { Radio } gives you a group control (single-select, options: { label }[]), and although RadioOption extends AriaRadioProps so the type looks like it accepts children, the component overwrites children with its own circle+label renderer, so only the label string ever shows — there's no slot for custom per-item content. That's the blocker here, because all three places this is used need exactly that: the dispute-feature picker (Court step) wraps each option in a help tooltip and renders a token-address TextField under the gated ERC-20/1155 options, the appeal OptionCard puts the indicator inside a card with a funding progress bar (its input is even display:none, so the card is the click target), and CreationCard is a rich icon/title/description card with a case-ID field. So this local component keeps the single-radio contract those cases need, rather than the group component.

web/src/pages/Cases/CaseDetails/Voting/VotingHistory.tsx
kemuru: if not needed externally condition doesn't hold here: currentTab feeds PendingVotesBox, VotesAccordion, the drawnJurors memo and isHiddenVotes, and the useEffect that auto-jumps to the latest round needs to set the selected tab, which only controlled selectedKey allows (defaultSelectedKey can't). So letting Tab own it would break the auto-latest-round behavior. 

did the equivalent simplification on CaseDetails/Tabs instead, where the value isn't needed externally.

web/src/styles/themes.ts
kemuru: the removal isn't the cause, that import had to go (v3 doesn't export the JS theme objects), and the spread was overridden by our local keys anyway, so app styles are unchanged. The real source is that v3 components are now themed by the library's theme.css CSS variables (imported in app.tsx), a separate channel from our styled-components theme.

web/src/styles/base-elements.css
```diff
@@ -0,0 +1,73 @@
...
+    color: var(--klerosUIComponentsSecondaryText);
+  }
+
+  a {
+    font-weight: 400;
+    font-size: 14px;
+    text-decoration: none;
+    color: var(--klerosUIComponentsPrimaryBlue);
+    transition: color 0.1s;
+  }
```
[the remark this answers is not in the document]
kemuru: pre-existing styling the PR only relocated (identical rule existed in global-style.ts on dev)

web/src/components/NumberInputField.tsx
```diff
@@ -1,14 +1,14 @@
 import React, { useState } from "react";
 import styled from "styled-components";
 
-import { Field } from "@kleros/ui-components-library";
+import { TextField } from "@kleros/ui-components-library";
```
kyrers, line 4: The library has `Number` fields. Shouldn't those be used here?

web/src/layout/Header/navbar/Menu/Settings/Notifications/FormContactDetails/index.tsx
```diff
@@ -145,7 +145,7 @@ const FormContactDetails: React.FC<ISettings> = ({ toggleIsSettingsOpen }) => {
       <ButtonContainer>
         <Button
           text={t("buttons.save")}
-          disabled={
+          isDisabled={
             !isEditingEmail || !emailIsValid || isAddingUser || isFetchingUser || isUpdatingUser || !isEmailUpdateable
           }
```
kyrers: This button doesn't seem to work for me. Maybe used to default to `type="submit"` and now doesn't?

web/src/layout/Header/navbar/Menu/Settings/index.tsx
```diff
@@ -63,8 +63,8 @@ const StyledTabs = styled(Tabs)`
 
 const Settings: React.FC<ISettings> = ({ toggleIsSettingsOpen, initialTab }) => {
   const { t } = useTranslation();
-  const [currentTab, setCurrentTab] = useState<number>(initialTab || 0);
   const containerRef = useRef(null);
+  const [currentTab, setCurrentTab] = useState<number>(initialTab || 0);
```
kyrers: Having the URL include `#notifications` does not open the settings modal for me. If that's the case, we can either fix it or remove the `currentTab` logic and let the library handle it. IMO, one of the two should be done.

web/src/layout/Header/navbar/Menu/Settings/index.tsx
```diff
@@ -74,30 +74,28 @@ const Settings: React.FC<ISettings> = ({ toggleIsSettingsOpen, initialTab }) =>
 
   const TABS = [
     {
+      id: 0,
       text: t("menu.general"),
       value: 0,
+      content: <General {...{ toggleIsSettingsOpen }} />,
```
kyrers: I have a bug where I see a scrollbar in this tab, and I don't think this happened before. 
That said, IIRC, I ran into this when migrating other apps. It was something having to do with the `Tabs` container.

web/src/pages/Cases/CaseDetails/Appeal/Classic/Options/StageTwo.tsx
```diff
@@ -51,17 +53,20 @@ const StageTwo: React.FC<IStageTwo> = ({ setAmount }) => {
           {fundedChoices.length > 0 && !choice.funded ? (
             <>
               <StageExplainer stage={2} countdown={winnerSideCountdown} />
-              <OptionsContainer>
+              <OptionsGroup
+                aria-label={t("appeal.which_option_to_fund")}
+                value={selectedOption?.id ?? null}
+                onChange={() => {}}
```
kyrers: Why the empty `onChange`? Before, it wasn't empty.

web/src/pages/Cases/CaseDetails/Appeal/Classic/Fund.tsx
```diff
@@ -122,21 +109,21 @@ const Fund: React.FC<IFund> = ({ amount, setAmount, setIsOpen, disputeKitName })
...
       <StyledLabel>{t("appeal.how_much_eth_contribute")}</StyledLabel>
-      <StyledField
+      <EthAmountField
         type="number"
         value={amount}
-        onChange={(e) => {
-          setAmount(e.target.value);
+        onChange={(value) => {
+          setAmount(value);
         }}
```
kyrers: Can be simplified to `onChange={setAmount}`. Simplifications like this are possible in other files as well. I didn't identify all of them to avoid spamming comments.

web/src/pages/Cases/CaseDetails/Appeal/EthAmountField.tsx
kyrers, line 1: The thing I mentioned before about the library having  Number and BigNumber fields applies here, too. Note that there are other places where this mismatch happens.

web/src/pages/Cases/CaseDetails/Appeal/EthAmountField.tsx
```diff
@@ -0,0 +1,32 @@
+import styled from "styled-components";
+
+import { TextField } from "@kleros/ui-components-library";
+
+/** Centered amount input with a static "ETH" suffix, shared by the Classic and Shutter appeal funding fields.
```
kyrers: I made this render, and I still don't see the ETH suffix. Might be missing `z-index`

web/src/pages/Cases/CaseDetails/Appeal/OptionCard.tsx
```diff
@@ -107,22 +97,28 @@ const OptionCard: React.FC<IOptionCard> = ({
...
-          </WinnerLabel>
-        </TextContainer>
-        {canBeSelected && <StyledRadio label="" checked={selected} />}
-      </TopContainer>
-      <LabelContainer>
-        <label>{fundingLabel}</label>
-      </LabelContainer>
-      <LinearProgress progress={progress} width={width} />
-    </StyledCard>
+    <StyledItem value={value} isDisabled={!canBeSelected}>
```
kyrers, line 158: So, before this was a `Card` with an optional `Radio` inside and now it's a ` CustomRadioItem` with a `Card` inside and an optional `RadioIndicator`? This forces `AppealHistory` and `StageTwo` to wrap their options in `CustomRadio` even if they are display only. Won't this cause accessibility issues?

web/src/pages/Cases/CaseDetails/Evidence/SubmitEvidenceModal.tsx
```diff
@@ -49,6 +49,21 @@ const EditorContainer = styled.div`
...
+  svg:has(+ [id="dropzone-label"]) {
+    fill: ${({ theme }) => theme.secondaryText};
+    path {
+      fill: ${({ theme }) => theme.secondaryText};
+    }
+  }
+  /* Align the icon to the first line of the message, not its vertical center. */
+  div:has(> [id="dropzone-label"]) {
+    align-items: flex-start;
+  }
```
kyrers, line 66: IMO, we should default to the library and avoid these customizations as much as possible. Otherwise, using the library won't result in all the apps having the same style. Also, defaulting to the library would not require us to use these selectors, which are fragile because if changed in the library, this breaks.
It's a matter of opinion, though.

web/src/pages/Cases/CaseDetails/Tabs.tsx
```diff
@@ -93,12 +103,10 @@ const Tabs: React.FC = () => {
 
   return (
     <StyledTabs
-      currentValue={currentTab}
+      selectedKey={currentTab < 0 ? 0 : currentTab}
+      defaultSelectedKey={currentTab < 0 ? 0 : currentTab}
```
kyrers: Seems there is unnecessary repetition here. Also, SonarCloud is right that this could be simplified with `Math.max`.

web/src/pages/Cases/CaseDetails/Tabs.tsx
kyrers, line 1: Using this file to make a more general point about `Tabs` usage:

In most usages I see, we're setting both the `selectedKey and defaultSelectedKey` and also not using the item `content` field to render the tab content. I understand this might be necessary in some situations, but I'm unsure if we can't use the `Tabs` component properly and simplify some usages.

web/src/pages/Resolver/Parameters/Court/FeatureSelection/index.tsx
```diff
@@ -161,21 +161,23 @@ const FeatureSelection: React.FC = () => {
         return (
           <Fragment key={groupName}>
             <GroupComponent clearAll={() => handleGroupDisable(groupName as Group)}>
-              <Fragment>
+              <CustomRadio
+                aria-label={groupName}
+                value={features.find((feature) => selected.includes(feature)) ?? null}
+                onChange={(value) => handleToggle(value as Features)}
```
kyrers, line 168: Before, clicking a radio button would unselect it, now we can't. I assume this is intentional?

web/src/pages/Resolver/Parameters/NotablePersons/PersonFields.tsx
```diff
@@ -103,16 +102,16 @@ const PersonFields: React.FC = () => {
             label={t("forms.labels.person_number", { number: index + 1 })}
             placeholder={t("forms.placeholders.alice_developer_example")}
             value={alias.name}
-            onChange={handleAliasesWrite}
+            onChange={(value) => handleAliasesWrite(index, "name", value)}
           />
           <LabeledInput
             name="address"
```
kyrers: Nice simplification in handling values! I think these `name` props are not even needed anymore. `OptionsFields` also has this simplification.

web/src/styles/themes.ts
kyrers, line 1: Can't this file be simplified? Seems like there are colors here that are already defined in the library theme. Even places where we do `color: theme.someColor` or similar, could probably use the var from the library directly, and it would update anyway on theme change, no? 
Would require refactoring in a lot of places, but to me it seems appropriate if we're migrating to the latest version of the library. Or can't this be done on `styled-components` and needs `tailwind` for some reason?

web/src/components/CasesDisplay/Filters.tsx
kyrers, line 1: Using this file to point out that there are lots of warnings throughout because `aria-labels` and related are not set.

web/src/pages/Cases/CaseDetails/Appeal/Classic/Options/StageTwo.tsx
kemuru: the reason is that on StageTwo there's only 1 possible selectable, so we automatically select it for the user instead. but i can improve this part of the code

web/src/pages/Resolver/Parameters/Court/FeatureSelection/index.tsx
kemuru: yeah

tractorss: We do need it to be unselectable, although this is not the appropriate behaviour for a radio component 🤔 

The requirement is such that when one feature is selected it disabled the non-compatible features in another group. So to make them selectable again, we need a way to unselect the former feature.
In this case, it was to just click again and deselect.

Can we do that with the new Radio component?

web/src/layout/Header/navbar/Menu/Settings/index.tsx
kemuru: yup nice one, actually i noticed this too but forgot to address it before

web/src/pages/Cases/CaseDetails/Evidence/SubmitEvidenceModal.tsx
kemuru: agree, but in the current state if we do that, multiple places don't match Figma's stylings

web/src/pages/Cases/CaseDetails/Tabs.tsx
kemuru: both quirks come from the wrapper Tabs: 

it destructures defaultSelectedKey out of props before forwarding to AriaTabs and keeps its own useState for the underline. So selectedKey controls the panel and defaultSelectedKey controls the underline. you need both. Fixable upstream.

content: null is structural: Profile, Description and this file are URL-driven, so the panels are owned by <Routes>. VotingHistory does the same with local state read by sibling components. Folding into items[].content means inlining Routes per item or rendering all panels. 

Settings is the one place the tab fully owns its content and we use content there. The right fix for the URL-driven ones is dropping library Tabs for nav and using a thin TabBar, maybe. in a different PR.

web/src/styles/themes.ts
kemuru: we should do it but honestly I would do it in a different PR, maybe in the tailwind one. ill start my vacation on tuesday so idk if we should keep this PR open more longer.

web/src/pages/Cases/CaseDetails/Appeal/OptionCard.tsx
kemuru: You were right. Added a selectable prop to OptionCard. When false, it skips the CustomRadioItem wrapper and renders just the Card. AppealHistory and StageTwo now use a plain <div role="list"> with selectable={false}, so no fake radio semantics around display-only content. StageOne still uses the default (selectable={true}) since that one's a real radio group.

Positive side effect: the empty onChange={() => {}} in StageTwo also went away since there's no RadioGroup left to feed, which clears up that other comment too.

web/src/components/NumberInputField.tsx
kemuru: Agreed, did this where it was clean: EthAmountField is now BigNumberField, Jurors.tsx is NumberField, and the ERC1155 token ID field in GatedErc1155 is BigNumberField.

But NumberInputField itself I left as TextField for now. Its two consumers (InputDisplay for PNK staking, CreationCard for dispute ID) both have string state wired into debounce + useParsedAmount that expects a ${number} template, plus a commify + roundNumberDown formatter that doesn't map onto the library's formatOptions (no flooring). BigNumberField also fires onChange(new BigNumber(0)) on empty input, which breaks CreationCard's val.trim() !== "" guard for "user cleared the field". So it's a coordinated consumer refactor, not a drop in replacement. Worth its own PR.

web/src/styles/base-elements.css
kyrers, line 1: I don't think this file is applying all styles properly. For instance, look at the General Court's policy, this should render like a list.

I also checked other policies and related, and `p` elements seem way too close. Maybe something to do with the margins.

Finally, I also see this line repeated across elements: `margin: 0 0 16px 0;`, doesn't this also influence elements coming from the ui-components-library?

Note that it's possible some of this is temporary and will go away with the migration to tailwind, I am not sure. You probably know what's going to stay or not.

Image

web/src/components/LabeledDropdown.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { useId } from "react";
+import styled from "styled-components";
+
+import { DropdownCascader, DropdownSelect } from "@kleros/ui-components-library";
+
+/**
+ * Accessible wrappers around the library's DropdownSelect / DropdownCascader.
+ *
+ * @kleros/ui-components-library 3.8.0 hardcodes `aria-label="Select"` after the prop spread,
```
kyrers: I assume this is the case in version `3.9.0` too. If I understand correctly, the problem here is that the library forces the `aria-label` to be `Select`, but isn't the fix in the `ui-components-library` itself? That way we could get rid of this entire wrapper, no? Not saying it must be done now, but I think I recall @tractorss saying we should improve the library where it makes sense, and this seems a quick win.

web/src/components/NumberInputField.tsx
```diff
@@ -1,14 +1,14 @@
...
 
 const Container = styled.div`
   width: 100%;
   height: fit-content;
 `;
 
-const StyledField = styled(Field)`
+const StyledField = styled(TextField)`
```
kyrers: I think the focus state regressed a bit. Not sure if: a) happens in other components and b) will be fixed in the migration to Tailwind.

Also, why is this a `TextField`? It should either be a `NumberField` or `BigNumberField`, no? 

Image

web/src/pages/Resolver/Parameters/Court/FeatureSelection/index.tsx
```diff
@@ -152,21 +152,23 @@ const FeatureSelection: React.FC = () => {
...
```
kyrers, line 158: As for the previous discussion here, clicking a selected option still doesn't unselect it.

web/src/pages/Resolver/Parameters/Jurors.tsx
```diff
@@ -79,24 +82,15 @@ const Jurors: React.FC = () => {
...
-
   return (
     <Container>
       <Header text={t("headers.select_number_of_jurors")} />
       <StyledField
         placeholder={t("forms.placeholders.select_the_number_of_jurors")}
-        value={noOfVotes}
-        onChange={handleJurorsWrite}
+        value={disputeData.numberOfJurors ?? NaN}
+        onChange={(numberOfJurors) => setDisputeData({ ...disputeData, numberOfJurors })}
```
kyrers: This no longer updates the arbitration cost as one types. You need to unfocus for the arbitration cost to update. Not sure how important this is, just mentioning the difference.

web/src/pages/Cases/CaseDetails/Tabs.tsx
```diff
@@ -92,14 +105,14 @@ const Tabs: React.FC = () => {
     return updatedTabs;
   }, [currentPeriodIndex, rounds.length, appealCost, TABS]);
 
+  // Both props: `selectedKey` keeps the URL the source of truth; `defaultSelectedKey`
+  // primes the library Tabs' internal underline state (which doesn't track `selectedKey`).
   return (
     <StyledTabs
-      currentValue={currentTab}
+      selectedKey={currentTab}
```
kyrers: The underline disconnects from the selected tab if the route changes without a click. The same applies to Profile and Courts description.

web/src/pages/Cases/CaseDetails/Timeline.tsx
```diff
@@ -37,10 +38,7 @@ const StyledSteps = styled(Steps)`
 
   h2 {
     font-size: ${responsiveSize(12, 14)};
-  }
-
-  [class*="horizontal-bullet__TextWrapper"] {
-    margin-top: 2px;
+    margin: 4px 0 0;
```
kyrers: The text is no longer aligned centrally with the icons.

web/src/pages/Cases/CaseDetails/Appeal/EthAmountField.tsx
```diff
@@ -0,0 +1,42 @@
...
+    top: 50%;
+    transform: translateY(-50%);
+    color: ${({ theme }) => theme.primaryText};
+    pointer-events: none;
+  }
+`;
+
+type Props = Omit<React.ComponentProps<typeof BigNumberField>, "onChange"> & {
+  onChange?: (value: string) => void;
+};
```
kyrers: There are more instances of this type of situation. I identified some and most likely missed others, but I don't think this is the correct way to migrate to the library. Why are we using `string` for a `BigNumberField`?

To me, the point of the migration is that consumers should change and adapt to the library, not us building wrappers around components just to avoid touching the consumers.

web/src/layout/Header/DesktopHeader.tsx
```diff
@@ -127,19 +127,21 @@ const DesktopHeader: React.FC = () => {
   const [isOnboardingMiniGuidesOpen, toggleIsOnboardingMiniGuidesOpen] = useToggle(false);
   const [initialTab, setInitialTab] = useState<number>(0);
   const location = useLocation();
+  const [searchParams] = useSearchParams();
   const navigate = useNavigate();
   const { isConnected, chainId } = useAccount();
   const isDefaultChain = chainId === DEFAULT_CHAIN.id;
+  // Notifications opens via `?notifications=true` (not a hash) so HashRouter doesn't swallow it.
+  const hasNotificationsPath = searchParams.get("notifications") === "true";
```
kyrers: Does any external app, like Atlas, use the previous link? If so, it'll stop working, no?

web/src/components/LanguageSelector.tsx
kyrers, line 1: There's a bug on smaller screens.

Image

web/src/layout/Header/navbar/Menu/Settings/Notifications/FormContactDetails/index.tsx
kyrers, line 1: Same visual bug as the Language Selector

Image

web/src/layout/Header/navbar/Menu/Settings/Notifications/FormContactDetails/index.tsx
```diff
@@ -203,23 +203,24 @@ const FormContactDetails: React.FC<ISettings> = ({ toggleIsSettingsOpen }) => {
                 event.preventDefault();
                 setIsConfirmingUnsubscribe(false);
               }}
-              disabled={isDeletingUser}
+              isDisabled={isDeletingUser}
             />
             <ConfirmUnsubscribeButton
               text={t("buttons.confirm_unsubscribe")}
               onClick={(event) => {
```
kyrers: Should be `onPress`. Also not sure if the `preventDefault` is needed.

kyrers (commented):
Some general notes:

- The console is still full of `If you do not provide a visible label, you must specify an aria-label or aria-labelledby attribute` warnings, mostly in the resolver.
- I created a case that had a link in the description, and I can't open it in the preview. Not sure if introduced in this PR.
- This [issue](https://github.com/kleros/kleros-v2/issues/2433) was meant to be solved during the migration. You can see it by trying to create a case and selecting "University" in the "Shielded voting" section. The icon does not render correctly, namely in dark mode.

kyrers:
@coderabbitai review

web/src/components/LabeledDropdown.tsx
tractorss: Hmm, agreed. I think in this case we can create the issue in ui component library for this. As this is not a huge blocker, in the sense not much of code is changing. 
We will ofc need to address this in the library itself.
wdyt

web/src/components/NumberInputField.tsx
tractorss: This can be a BigNumberField agreed, unless there's a reason not too.
Try to avoid NumberField for any input related to amounts that go on-chain.

web/src/pages/Resolver/Parameters/Court/FeatureSelection/index.tsx
kemuru: there's a button for clearing the selection. i would argue its weird to unselect it by clicking it again

web/src/layout/Header/DesktopHeader.tsx
kemuru: Atlas documents /#/settings/notifications, which never existed. just added a redirect route so it works

web/src/components/LabeledDropdown.tsx
kemuru: just created the issue at https://github.com/kleros/ui-components-library/issues/96

web/src/hooks/useBigNumberFieldReset.ts
kyrers, line 1: This workaround does work, but again, it seems like something that should be fixed in the ui-component-library itself. Not sure if now, but at least worthy of an issue there, in my opinion. 
We must also keep track of this type of thing so we know what needs changing after the ui-component-library is fixed.

web/src/pages/Courts/CourtDetails/StakePanel/InputDisplay.tsx
```diff
@@ -5,24 +5,37 @@ import { useTranslation } from "react-i18next";
...
+    &:focus {
+      position: relative;
+      z-index: 1;
+    }
+  }
+
+  /* Hover-revealed stepper arrows would sit under the button's edge. */
+  & .input-wrapper > div:has(> button[aria-label="Increment"]) {
+    display: none;
   }
```
kyrers: I've noticed this in other files too. It seems fragile if the DOM in the ui-component-library changes. Probably, the better solution is to add a prop to hide this in the library itself. I would create an issue for this.

web/src/pages/Resolver/Landing/CreationCard.tsx
```diff
@@ -100,37 +100,50 @@ interface ICreationCard {
...
               placeholder={t("forms.placeholders.case_id_example")}
               value={disputeID}
-              onChange={(val) => {
-                if (setDisputeID) setDisputeID(val.trim() !== "" ? val : undefined);
-              }}
+              // The library reports an emptied input as 0, and 0 is a real dispute; read the
+              // raw text to tell "cleared" apart from "0".
+              onChange={(id) =>
+                setDisputeID?.(id.isZero() && !disputeIdInputRef.current?.value.trim() ? undefined : id.toString())
+              }
```
kyrers, line 141: Hmm, maybe another issue in the library for this? I am not sure if there's a reason for 0 to be returned when the input is empty. Maybe it could even be folded into the same issue I suggested in the `useBigNumberFieldReset` hook.

web/src/components/DisputeFeatures/Features/FeatureRadio.tsx
kyrers, line 1: I don't know of another way of having the radio buttons in the app be deselected by clicking them again, other than adding the feature ourselves, like you're doing here. However, if we really want this, it probably should be a feature in the ui-component-library, no? It probably can even be done in a cleaner way there.

web/src/components/DisputeFeatures/Features/index.tsx
```diff
@@ -1,48 +1,32 @@
...
-};
-
-export type FeatureUI = React.FC<RadioInput>;
-
-export const StyledRadio = styled(Radio)`
-  font-size: 14px;
-  color: ${({ theme, disabled }) => (disabled ? theme.secondaryText : theme.primaryText)};
-  opacity: ${({ disabled }) => (disabled ? "0.7" : 1)};
-`;
+export { FeatureRadio, type FeatureUI, type RadioInput } from "./FeatureRadio";
```
kyrers: Seems unnecessary, as the other files will still import from `"./"`. This pattern is useful for barrel index files, and I don't think this is a pattern we have in other places in the app.

web/src/components/LabeledDropdown.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { DropdownCascader, DropdownSelect } from "@kleros/ui-components-library";
+
+/**
+ * Accessible wrappers around the library's DropdownSelect / DropdownCascader.
+ *
+ * @kleros/ui-components-library 3.9.0 hardcodes `aria-label="Select"` after the prop
+ * spread, silently dropping caller-supplied `aria-label`. Workaround: pass `aria-labelledby` (which
+ * the spread does forward) pointing to a visually-hidden span; per ARIA, labelledby wins.
+ * Tracked in kleros/ui-components-library#91; once fixed there, these wrappers go away.
```
kyrers: You mentioned [issue 96 ](https://github.com/kleros/ui-components-library/issues/96) in the thread comment, so seems there are two issues for the same thing.

kyrers (commented):
Also, retracting my comment regarding links in the description not opening during the preview step. Seems that is done intentionally, as there's a div blocking interactions.

web/src/pages/Courts/CourtDetails/StakePanel/InputDisplay.tsx
tractorss: Why do we need to hide the stepper buttons here?

I tried reading the comments, but from the comment alone I am not able to get the idea still

web/src/components/DisputeFeatures/Features/FeatureRadio.tsx
tractorss: So when I was adding the Feature selection feature, that time Radio unselection was available, but even then Radio options aren't usually "un-selectable" , an option has to be selected.

Can we instead add a new , default, "Clear" or "None" option in each Group?

The unselection is needed so the other pairs can become available to be selected again and I think introducing "None" an option should act the same way and we can be more explicit about it 🤔

web/src/layout/Header/DesktopHeader.tsx
tractorss: Please share this with the atlas team instead of silent redirect 🙏

web/src/pages/Cases/CaseDetails/Appeal/EthAmountField.tsx
tractorss: Was this addressed? Please don't resolve without a reply, it's hard to see if this was addressed or not 🙏

kemuru: Yup. Here's a summary from Claude, I read it and I agree:

- `EthAmountField.tsx` is now `styled(BigNumberField)` and nothing else. The `Props` type and the `onChange` string adapter are removed.

- Its consumer, `Appeal/Classic/Fund.tsx`, uses the library's API directly:
  `value={amount || undefined}` and `onChange={(value) => setAmount(value.isZero() ? "" : value.toString())}`.

- The other wrapper of this kind, `NumberInputField.tsx` (a `TextField` wrapper used for the stake amount and the case ID), is deleted. `StakePanel/InputDisplay.tsx` and `Resolver/Landing/CreationCard.tsx` use `BigNumberField` directly.

- On "why string": the value goes to viem's `parseUnits(amount, 18)`, which takes a decimal string, so the conversion happens in the consumer that owns the state. Keeping a `BigNumber` in state does not work with 3.9.0: `useBigNumberField` has `useEffect(() => { if (value !== undefined) setInputValue(value.toString()) … }, [value])` (lines 169–171). A `BigNumber` state is a new object on every `onChange`, so the effect rewrites the input each keystroke: typing `1.` renders `1`, and typing `5` next gives `15`. With a string, `1.` and `1` both map to `"1"`, the prop is unchanged, and typing `1.5` works. Both cases checked against the library in a test.

- The `isZero() ? "" : …` mapping is there because the library calls `onChange(new BigNumber(0))` when the input is emptied (`useBigNumberField.tsx` lines 355 and 390). That is part of the library issue "BigNumberField: controlled mode cannot represent an empty field"; the two consumer sites carry a `TODO(ui-components-library)` tag to drop the mapping once it ships.

web/src/layout/Header/DesktopHeader.tsx
kemuru: Done

web/src/components/DisputeFeatures/Features/FeatureRadio.tsx
kyrers: Yeah, I suggested this because when you said in another thread that it needs to be unselectable, I understood it as being unselectable by clicking again. My bad. I do agree this is not normal behaviour for radio buttons, so I am fine with it going away.

web/src/components/Verdict/DisputeTimeline.tsx
```diff
@@ -34,21 +35,32 @@ const Container = styled.div`
 
 const StyledTimeline = styled(CustomTimeline)`
   width: 100%;
+
+  /* TODO(ui-components-library): remove once CustomTimeline centers its bullet on the title
+     (issue: "CustomTimeline: bullet is top-aligned with the title instead of centered").
```
kyrers: I don't see this issue in the library 🤔

web/src/components/DisputeFeatures/Features/FeatureRadio.tsx
kemuru: There is a button called "Clear" in the feature selection for the dispute kits, which is why I didn't mind to not have the unselect behavior. I only added it because of the suggestion, but I was fine with not having it, since yeah, one of them needs to be selected anyways. Nevertheless we can keep the status quo and worry about updating this later since there will be an update in ui-components-library.

web/src/components/Verdict/DisputeTimeline.tsx
kemuru: true. i didn't create one for this. done, created as issue number 100

web/src/pages/Courts/CourtDetails/StakePanel/InputDisplay.tsx
kemuru: I guess the only justification was to match dev. But I would be fine with showing them here.

kemuru: Just pushed showing them here for now.

tractorss: > I guess the only justification was to match dev. But I would be fine with showing them here.

We don't want to mold the library into the exact pre-migration UI in court v2.
Some features were added in the ui components library v3, and hence why we are updating. Explicitly removing functionalities like this just to match the current state of dev should not be the first step.
If something is breaking or changing the behvaiour alot, we can discuss

web/src/components/DisputeFeatures/Features/FeatureRadio.tsx
tractorss: > I understood it as being unselectable by clicking again.

Yes with v2 version of the libray , where unselecting a Radio component was possible, this was what we had. Even then we had doubts about making the option unselectable, but it was urgently needed at the time. We had this discussion with Aleix at the time.

But now with the updated library and the RadioGroup component , I think it's better to make the "None" option.

As Marino pointed, "Clear" achieves the same. So for the time being we can have that and not use "None" option

kyrers: I am not clear here. Is the intention to keep the behavior for now, or remove it right away? If we want it removed right away, can't the whole deselect logic be removed here and from the tests?

web/src/styles/commonStyles.ts
```diff
@@ -7,3 +7,52 @@ export const hoverShortTransitionTiming = css`
...
+      fill: ${({ theme }) => theme.stroke};
+    }
+  }
+`;
+
+/* TODO(ui-components-library): remove once NumberField/BigNumberField take a `hideStepper` prop
+   (kleros/ui-components-library#98).
+   Hides the hover-revealed ±1 arrows of a BigNumberField where stepping makes no sense (an ID)
+   or where they would overlap a suffix drawn in the input (the ETH amount). Relies on the library's DOM:
+   the arrows sit in the first div of `.input-wrapper` and carry a fixed English aria-label. */
```
kyrers, line 53: Following from the other thread about steppers, I think some discussion is needed here.
Personally, I do think steppers make sense in the ID. I also think that the overlap is caused on the app side, so the fix should be here.

web/src/components/DisputeFeatures/Features/FeatureRadio.tsx
tractorss: I'll move this to slack, and put the context there, so we can discuss properly

web/src/styles/commonStyles.ts
tractorss: I also think it makes sense to have , even for Id

tractorss: Apart from that , I also don't see any other reason for the steppers to not be available when the input requires a number input. 
If it's a read-only field, we should disable the field instead in that case and stepper won't show

tractorss: let's remove this for now, but discuss the need for stepper removal ltr. We can add it to ui-components library if the impact is minimal on stepper removal

@coderabbitai mention the learnings from this thread on the issue #98 on ui components library repo

PierreUntas:
I tested the deploy preview for this PR on the Courts page
(#/courts/1/purpose) and the top search dropdown behaves the same
as production for me:

- ArrowDown / ArrowUp: no highlight, no movement
- Enter: nothing, even with a single exact match
- Escape: nothing, the dropdown stays open
- Click outside: nothing, the dropdown stays open (clearing the
  input is the only way to close it)

Preview: https://deploy-preview-2281--kleros-v2-testnet.netlify.app/#/courts/1/purpose

Context: #1823 is still open, and #2106 was closed in January with
"Should be done using the new UI components library". TopSearch.tsx
is listed in the changed files here, so I wanted to check. Is this
dropdown in the migration scope, or is it still using the old
component? If the latter, where would be the right place to track
this: here, #1823, or an issue in ui-components-library?

